PR #6, opened 2022-03-21
Unify the IntelliJ plugin version 

DanielJette:

Previously the IntelliJ plugin was a separate gradle projected. This PR changes the gradle structure so that the IntelliJ plugin is just another module in the same gradle project graph as all the other modules. The main advantage of this is that it unifies the version number (currently 1.2.0-alpha01) which had previously been using an independent versioning.

### Tophat

- Checkout this branch and run `./gradlew IntelliJ:buildPlugin` from the root
- The built plugin can be found at `/.android-testify/Plugins/IntelliJ/build/libs/IntelliJ-1.2.0-alpha01.jar`

### DIFF
--- a/Plugins/IntelliJ/gradle.properties
+++ b/Plugins/IntelliJ/gradle.properties
@@ -3,9 +3,9 @@
 
 pluginGroup = dev.testify
 pluginName = Android Testify - Screenshot Instrumentation Tests
-pluginVersion = 0.1.2
 pluginSinceBuild = 193
 pluginUntilBuild = 211.*
+javaVersion = 1.8
 
 platformType = IC
 platformVersion = 2020.1
--- a/build.gradle
+++ b/build.gradle
@@ -3,9 +3,7 @@ buildscript {
         mavenLocal()
         google()
         mavenCentral()
-        maven {
-            url "https://plugins.gradle.org/m2/"
-        }
+        gradlePluginPortal()
     }
 
     ext {
--- a/build.gradle
+++ b/build.gradle
@@ -3,6 +3,9 @@ buildscript {
         mavenLocal()
         google()
         mavenCentral()
+        maven {
+            url "https://plugins.gradle.org/m2/"
+        }
     }
 
     ext {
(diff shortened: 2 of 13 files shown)


Plugins/IntelliJ/gradle.properties
```diff
@@ -3,9 +3,9 @@
 
 pluginGroup = com.shopify.testify
 pluginName = Android Testify - Screenshot Instrumentation Tests
-pluginVersion = 0.1.2
 pluginSinceBuild = 193
 pluginUntilBuild = 211.*
+javaVersion = 1.8
```
jaredh, line 8: Do we actually have to build the plugin with Java 8? Everything else is on 11 now.

Plugins/IntelliJ/gradle.properties
```diff
@@ -3,9 +3,9 @@
...
 
 platformType = IC
 platformVersion = 2020.1
```
jaredh, line 11: Looks like this version has problems building on `aarch64`. I'll log an issue for looking adding M1 support.

build.gradle
```diff
@@ -3,6 +3,9 @@ buildscript {
         mavenLocal()
         google()
         mavenCentral()
+        maven {
+            url "https://plugins.gradle.org/m2/"
```
jaredh, line 7: I think `gradlePluginPortal()` resolves to this?

jaredh (approved):
Had problems resolving platform specific dependencies for `org.jetbrains.jbre` but I bumped the `platformVersion` to a very recent IntelliJ release to test building this successfully on `arm64`. After that, 👌.

Plugins/IntelliJ/gradle.properties
DanielJette: I added this in response to a build error. I suspect that, yes, it requires 8.
There's a bunch of places that explicitly request java 8. 🤷 

```
./Plugins/IntelliJ/build.gradle.kts:59: sourceCompatibility = "1.8"
./Plugins/IntelliJ/build.gradle.kts:60: targetCompatibility = "1.8"
./Plugins/IntelliJ/build.gradle.kts:64: kotlinOptions.jvmTarget = "1.8"
```

build.gradle
DanielJette: Yeah, I think you're right. I might have looked at some old documentation. I'll try changing this.

DanielJette: Changed

PR #25, opened 2022-04-05
Replace com.shopify.testify with dev.testify in markdown files

DanielJette:

Replace `com.shopify.testify` with `dev.testify` in the markdown files.

You can view the rendered markdown files here:

- [CODE_OF_CONDUCT.md](https://github.com/ndtp/android-testify/blob/update-strings/CODE_OF_CONDUCT.md)
- [Ext/Compose/README.md](https://github.com/ndtp/android-testify/blob/update-strings/Ext/Compose/README.md)
- [README.md](https://github.com/ndtp/android-testify/blob/update-strings/README.md)
- [RECIPES.md](https://github.com/ndtp/android-testify/blob/update-strings/RECIPES.md)

This PR also updates the Gradle Plugin to reference `dev.testify` in some strings

### DIFF
--- a/Ext/Compose/README.md
+++ b/Ext/Compose/README.md
@@ -2,7 +2,7 @@
 
 Easily create screenshot tests for `@Composable` functions.
 
-<a href="https://search.maven.org/artifact/com.shopify.testify/testify-compose"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/com.shopify.testify/testify-compose?color=%236e40ed&label=com.shopify.testify%3Atestify-compose"/></a>
+<a href="https://search.maven.org/artifact/dev.testify/testify-compose"><img alt="Maven Central" src="https://img.shields.io/maven-central/v/dev.testify/testify-compose?color=%236e40ed&label=dev.testify%3Atestify-compose"/></a>
 
 ---
 
@@ -15,7 +15,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath "com.shopify.testify:plugin:1.2.0-alpha01"
+        classpath "dev.testify:plugin:1.2.0-alpha01"
     }
 }
 ```
@@ -23,7 +23,7 @@ buildscript {
 **Application build.gradle**
 ```groovy
 dependencies {
-    androidTestImplementation "com.shopify.testify:testify-compose:1.2.0-alpha01"
+    androidTestImplementation "dev.testify:testify-compose:1.2.0-alpha01"
 }
 ```
 
@@ -55,7 +55,7 @@ class ComposableScreenshotTest {
 
     MIT License
     
-    Copyright (c) 2021 Shopify
+    Copyright (c) 2022 ndtp
     
     Permission is hereby granted, free of charge, to any person obtaining a copy
     of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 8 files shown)


Ext/Compose/README.md
```diff
@@ -55,7 +55,7 @@ class ComposableScreenshotTest {
 
     MIT License
     
-    Copyright (c) 2021 Shopify
+    Copyright (c) 2022 ndtp
```
jaredh, line 58: Copyright  on original works still must be attributed to Shopify prior to the fork. We'll need to represent this throughout with something like this.

```suggestion
    Copyright (c) 2022 ndtp
    Copyright (c) 2018-2022 Shopify
```

DanielJette: I looked up a guide on copyright and it implied that, once you change a file, the old copyright is invalidated.
I don't mind keeping the Shopify copyright, but at what point is it a new file? Everything has been renamed, so it's technically not the same file anymore.
Do you have any references on how to handle this? I'll see if I can dig up the document I read.

@jaredh

jaredh: IANAL but the [MIT license](https://opensource.org/licenses/MIT) states

> Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated documentation files (the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
> 
> **The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.**

I think headers are fine to scrap in general as long as attribution remains in `LICENSE`.

DanielJette: None of this is very clear to me. I'll redo/change/update anything here that I need to. I don't want to get any lawyers after us, but man, it's confusing to me.

See, like I read that and think "substantial portions" means that, once you cross a threshold, it doesn't matter anymore ...
But it could also be a single function that remains unchanged. I'd rather be safe-than-sorry, but I also don't want to attribute ownership to Shopify if they shouldn't have it.

I'm thinking changing everything to ` Copyright (c) 2022 ndtp, Copyright (c) 2018-2022 Shopify` for any files that were migrated over from the old repo and use only `Copyright (c) 2022 ndtp` for any entirely new files.

@jaredh

jaredh: This [SE discussion](https://softwareengineering.stackexchange.com/questions/157968/how-to-manage-a-copyright-notice-in-an-open-source-project) has a nice suggestion of differentiating with "Original work" and "Modified work".

> I'm thinking changing everything to Copyright (c) 2022 ndtp, Copyright (c) 2018-2022 Shopify for any files that were migrated over from the old repo and use only Copyright (c) 2022 ndtp for any entirely new files.

+1, I think that's the safest approach.

DanielJette: Really like that SO suggestion. I think I'll do that. I can't see any problem with it

For any migrated/modified files, I'll use:

```
Original work Copyright (c) 2018 Shopify 
Modified work Copyright (c) 2022 ndtp
```

For new files, I'll use:
```
Copyright (c) 2022 ndtp
```

And I'll basically double the LICENSE.md file to include both an ndtp section and a Shopify section.

@jaredh

DanielJette: I made the changes here https://github.com/ndtp/android-testify/pull/30

I'll re-do this PR once #30 is merged.

PR #35, opened 2022-04-13
Use the project name for the IntelliJ build/distributions filename

DanielJette:

Should fix the [error](https://github.com/ndtp/android-testify/runs/5996463067?check_suite_focus=true) on `upload-release-asset` currently on `main`

`Error: ENOENT: no such file or directory, stat './plugin-artifact/IntelliJ-1.2.0-alpha01.zip'`

This is caused because the `intellij_build.xml` actions expect the artifact name to be `<project>-<version>.zip` but we were setting the build artifact name to be `<custom project name>-<version>.zip`. So, I removed the `properties("pluginName")` and use the `project.name` instead.


jaredh (approved):
That makes more sense 🙂

PR #37, opened 2022-04-20
Issue-8, Issue-33: Add support for Android Studio Chipmunk and Dolphin

### LINKED ISSUE #8: Add IntelliJ plugin support for Android Studio Chipmunk (2021.2.1 Beta 4)
**Is your feature request related to a problem? Please describe.**

This feedback relates to:
- [ ] The Kotlin library
- [ ] The Gradle plugin
- [x] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**Describe the solution you'd like**

Chipmunk is available here https://developer.android.com/studio/preview and it should be supported.

**Describe alternatives you've considered**

**Additional context**

DanielJette:

Resolves #8, #33 

### How have you achieved it?

The main bug was that the `ktFile` module name now includes the test variant (`androidTest`) as a suffix. This needs to be trimmed to match the expected Gradle project name.

However, I had to make a lot of changes to the build.gradle file for the IntelliJ plugin because bumping the target IntelliJ version started building against newer IC library versions and pretty much nothing compiled. So, I refactored the build to use explicit variables for the compile and run instances of Android Studio. This makes it easy to build against a fixed version of the IntelliJ platform while testing against any given installation of Android Studio.

### Tophat instructions

1. Check out this branch `bugfix/33/android-studio-213`
2. Download Dolphin or Chipmunk from https://developer.android.com/studio/preview
3. Edit `./Plugins/IntelliJ/gradle.properties` and change the `StudioRunPath` to point to your Android Studio installation from step 2
4. Run `./gradlew IntelliJ:buildPlugin IntelliJ:runIde`
5. Open the `Sample` source code in the preview build of Android Studio
6. Verify that the 📷 icon is present in the screenshot test files and that the various commands work as expected.

Or, you can install the pre-built plugin from [IntelliJ-1.2.0-alpha01.jar.zip](https://github.com/ndtp/android-testify/files/8523615/IntelliJ-1.2.0-alpha01.jar.zip)

### DIFF
--- a/Plugins/IntelliJ/build.gradle.kts
+++ b/Plugins/IntelliJ/build.gradle.kts
@@ -6,18 +6,24 @@ import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 
 fun properties(key: String) = project.findProperty(key).toString()
 
+@Suppress("UNCHECKED_CAST")
 val versions = rootProject.extra["versions"] as Map<String, String>
 val testifyVersion = versions["testify"] as String
 
+if (!hasProperty("StudioCompilePath")) throw GradleException("No StudioCompilePath value was set. Please define StudioCompilePath in gradle.properties.")
 plugins {
     id("org.jetbrains.kotlin.jvm")
-    id("org.jetbrains.intellij")
-    id("org.jetbrains.changelog")
-    id("org.jlleitschuh.gradle.ktlint")
+    id("org.jetbrains.intellij") version "1.4.0"
+    id("org.jetbrains.changelog") version "1.3.1"
+    id("org.jlleitschuh.gradle.ktlint") version "10.2.1"
 }
 
-group = properties("pluginGroup")
+group = project.property("pluginGroup") as String
 version = testifyVersion
+
+val studioRunPath = project.property("StudioRunPath") as String
+val studioCompilePath = project.property("StudioCompilePath") as String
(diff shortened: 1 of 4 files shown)


DanielJette:
Oh hmm ... 

```
Specified localPath '/Applications/Android Studio Preview 2021.3.1.8.app/Contents' doesn't exist or is not a directory
```

Yeah, I guess we don't have Android Studio installed on CI 🤦 

I don't think we need that, but I'll need to tweak some stuff

DanielJette:
> Oh hmm ...
> 
> ```
> Specified localPath '/Applications/Android Studio Preview 2021.3.1.8.app/Contents' doesn't exist or is not a directory
> ```
> 
> Yeah, I guess we don't have Android Studio installed on CI 🤦
> 
> I don't think we need that, but I'll need to tweak some stuff

Fixed this. The problem is that there are two ways to define the source for the intellij plugin, either with `intellij.version` (which takes a platform number) or the `intellij.localPath` property (which takes a path to a specific installation).
In general, we want to build against a known version, not a specific local installation path. So, for CI, we use `intellij.version`. But, for local debugging purposes, you can define a `StudioRunPath` variable to point towards a specific installation.

jaredh (approved):
🎩'd on aarc64 and looks 👌

PR #40, opened 2022-04-22
Refactor ScreenshotRule to abstract out the configuration from the JUnit4 overrides

DanielJette:

Refactor ScreenshotRule to abstract out the configuration from the JUnit4 overrides

### How have you achieved it?

This is a preliminary refactoring of `ScreenshotRule` to make its functionality more generic. Eventually, this functionality will be promoted up to a `TestifyCore` class. For now, these changes make the `ScreenshotRule` rely less on `JUnit4` and exposes the configuration and modification methods in a more generic manner.

See https://github.com/ndtp/android-testify/blob/main/docs/architecture/library.md#screenshotextension

### Tophat instructions

No functionality or public API changes. All tests should pass as before.

### Notice


jaredh (approved):
LGTM.

PR #64, opened 2022-05-31
Issue-61: Part 1 - Create accessibility extension module

DanielJette:

Adds the scaffolding to create a new `:Accessibility` module.

### How have you achieved it?

This is just the boilerplate. There is no code yet.

### Tophat instructions

`./gradlew :Accessibility:assembleDebug` to verify that it builds.

### Notice

### DIFF
--- a/Ext/Accessibility/build.gradle
+++ b/Ext/Accessibility/build.gradle
@@ -68,14 +68,13 @@ android {
     dependencies {
         implementation project(":Library")
 
-        implementation 'com.google.code.gson:gson:2.9.0'
         implementation "androidx.core:core-ktx:${versions.androidx.core}"
-        implementation "androidx.test:rules:${versions.androidx.test.rules}"
-        implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk8:${versions.kotlin}"
-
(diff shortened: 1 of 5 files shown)


Ext/Accessibility/build.gradle
```diff
@@ -0,0 +1,91 @@
...
+    kotlinOptions {
+        jvmTarget = "1.8"
+        useIR = true
+    }
+
+    dependencies {
+        implementation project(":Library")
+
+        implementation 'com.google.code.gson:gson:2.9.0'
+        implementation "androidx.core:core-ktx:${versions.androidx.core}"
```
gantonious, line 72: Is the fix of hardcoded + centralized versions intentional?

gantonious (approved):

Ext/Accessibility/build.gradle
DanielJette: Ah, no. That's a mistake. Thanks for catching that @gantonious 

The Sample app is the only project that should be using direct-references to versions. All the other libraries I want a centralized place to update everything, so this should be using the `version` variables.

I'll fix this before I merge it

PR #65, opened 2022-05-31
Issue-63: Part 2 - Create screenshot lifecycle extensions

DanielJette:

Create screenshot lifecycle extensions

### How have you achieved it?

There already existed 5 lifecycle methods that were exposed via `ScreenshotRule` that could be used by sublcasses of `ScreenshotRule` to hook into certain points of the screenshot test. These methods are `beforeAssertSame`, `beforeInitializeView`, `afterInitializeView`, `beforeScreenshot`, and `afterScreenshot`. The idea was to allow users to run custom code at given moments while the test was running.

This PR takes those 5 methods and pulls them into an interface: `ScreenshotLifecycle`. Then, I added `addScreenshotObserver`/`removeScreenshotObserver` methods so that anyone could subscribe to these events, not just subclasses. I refactored `ScreenshotRule` so that the public API remains the same and subclasses can continue to hook into those methods.

This means that it's now much easier to add custom behaviour to screenshot tests. This will be used by the Accessibility checks in future PRs.

⭐ - I have also included a couple of minor changes to the visibility of some methods which will be handy for future work on the Accessibility module.

### Tophat instructions

- All tests should pass

### Notice


Library/src/main/java/dev/testify/ScreenshotRule.kt
```diff
@@ -587,9 +570,9 @@ open class ScreenshotRule<T : Activity> @JvmOverloads constructor(
                     return
                 }
 
-                beforeInitializeView(activity)
+                screenshotLifecycleObservers.map { it.beforeInitializeView(activity) }
```
gantonious, line 573: Can use `forEach` instead since we don't care about the result.

Library/src/main/java/dev/testify/ScreenshotRule.kt
```diff
@@ -444,29 +448,13 @@ open class ScreenshotRule<T : Activity> @JvmOverloads constructor(
...
-     * Invoked prior to any view modifications and prior to layout inflation.
-     */
-    open fun beforeInitializeView(activity: Activity) {}
-
-    /**
-     * Test lifecycle method.
-     * Invoked after layout inflation and all view modifications have been applied.
-     */
-    open fun afterInitializeView(activity: Activity) {}
+    fun addScreenshotObserver(observer: ScreenshotLifecycle) {
```
gantonious, line 451: Non blocking thought, make this could be part of the configuration? Allows for scoping observers to specific tests?

gantonious (approved):

Library/src/main/java/dev/testify/ScreenshotRule.kt
DanielJette: Yes, good call.

Library/src/main/java/dev/testify/ScreenshotRule.kt
DanielJette: 🤔 
Very interesting idea. I'll think on the implications of this kind of change, but it might be cleaner & more flexible if it's part of the configuration.

gantonious: I'm also not 100% familiar with how junit executes test cases. Do we get a new rule for each test case or does it get shared across the suite?

DanielJette: It's shared within the class. You get a new instance for each test class instance.

PR #66, opened 2022-05-31
Issue-63: Part 3 - Implement accessibility checks

DanielJette:

To help people with disabilities access Android apps, developers of those apps need to consider how their apps will be presented to accessibility services. Some good practices can be checked by automated tools, such as if a View has a contentDescription. Other rules require human judgment, such as whether or not a contentDescription makes sense to all users. 

Enabling `assertAccessibility` on `ScreenshotRule` will run the latest set of checks as defined by the [Accessibility Test Framework for Android](https://github.com/google/Accessibility-Test-Framework-for-Android). This library collects various accessibility-related checks on `View` objects as well as `AccessibilityNodeInfo` objects (which the Android framework derives from Views and sends to AccessibilityServices). 

See https://developer.android.com/training/testing/espresso/accessibility-checking

All elements within the hierarchy defined by `android.R.id.content` will be available to the check.

To enable accessibility checks:

1. Add a dependency to the Accessibility checks library `androidTestImplementation "dev.testify:testify-accessibility:1.2.0-alpha01"`
2. Invoke the `assertAccessibility` extension method on `ScreenshotRule`
    ```kotlin
    @ScreenshotInstrumentation
    @Test
    fun default() {
        rule
            .assertAccessibility()
            .assertSame()
    }
    ```

### Tophat instructions

- Tests will be added in a future PR
- For now, you can verify that this builds: `./gradlew Accessibility:assembleDebug`

### Notice


Ext/Accessibility/src/main/java/dev/testify/accessibility/internal/BaselineGenerator.kt
```diff
@@ -0,0 +1,77 @@
...
+/**
+ * Read the a11y.json file for the current test.
+ */
+internal fun readAccessibilityCheckResults(context: Context): CheckResults? {
+    val description = getInstrumentation().testDescription
+    val filePath = OutputFileUtility().getFileRelativeToRoot(
+        subpath = DeviceIdentifier.getDescription(context),
+        fileName = description.name,
+        extension = JSON_EXTENSION
+    )
```
gantonious, line 55: I'm a bit confused why the file resolution is different for the read and write cases? Any reason we couldn't do:

```kt
val rawJson = getOutputFile(context, description).readText()
```

Ext/Accessibility/src/main/java/dev/testify/accessibility/internal/AccessibilityScreenshotLifecycleObserver.kt
```diff
@@ -0,0 +1,101 @@
...
+    }
+
+    private fun AccessibilityHierarchyCheckResult.toCheckResult(): CheckResult {
+        return CheckResult(
+            id = this.resultId,
+            type = this.type.name,
+            check = this.sourceCheckClass.simpleName,
+            elementClass = this.element?.className.toString(),
+            resourceName = this.element?.resourceName.toString(),
+            message = this.getMessage(Locale.getDefault()).toString()
```
gantonious, line 98: Hmm would it make sense to use `Locale.ROOT` here? Not sure how this would behave if the device was using a non EN language?

gantonious (approved):
Overall looks good, just a couple of questions.

Ext/Accessibility/src/main/java/dev/testify/accessibility/internal/BaselineGenerator.kt
DanielJette: Oh, yes. The reason this is different is because we `read` from the `assets` directory packaged in & deployed with the APK.
For the `write`, we write the file to `data/data` (or the sdcard, depending on configuration).

We can't/don't want to write anything to `assets` as those are the pristine baselines.

@gantonious

Ext/Accessibility/src/main/java/dev/testify/accessibility/internal/AccessibilityScreenshotLifecycleObserver.kt
DanielJette: 🤔 
I wonder if the error messages are localized at all.
I'll try with ROOT though as that could be "safer"; these should be stable for all users.

PR #69, opened 2022-06-04
Add compose glide image test to sample

DanielJette:
⚠️ Migrate from https://github.com/Shopify/android-testify/pull/265

- Added the https://github.com/skydoves/landscapist/releases library for using Glide in compose.
- Added a very simple `ImageDemo` that displays `avatar1` (see the image above the list in screenshot that it works when testing it in `ComposeActivity`)
    
- Added screenshot test for the image in `ComposableScreenshotTest` attempting to use `CountingIdlingResource` to wait for `onResourceReady`


Sample/src/androidTest/java/dev/testify/sample/compose/glide/GlideImageScreenshotTest.kt
```diff
@@ -0,0 +1,63 @@
...
+        rule
+            .setCompose {
+                Box(
+                    modifier = Modifier
+                        .fillMaxWidth()
+                        .wrapContentHeight()
+                ) {
+                    GlideImageDemo(onImageLoaded = { idlingResource.decrement() })
+                }
+                IdlingRegistry.getInstance().register(idlingResource)
```
gantonious, line 59: I don't know compose very well yet. Is it safe to run non composable code like this in a composable block?

DanielJette: Oh, you know, I was so eager to get this working that I didn't consider the composition lifecycle. It's allowed to run non-compose code in a composable block, but it needs to be done in such a way that it safely handle recomposition.
I'm going to take this offline and look more closely at this.
My gut says that this will get hammered by recompositions and that would be bad.

Thanks for the questions @gantonious

Issue #99, opened 2022-10-11
[ComposableScreenshotRule] API 33 -- android.content.ActivityNotFoundException: Unable to find explicit activity class 


djette-st:
**Describe the bug**

As of API 33, you can no longer run screenshot test that use `ComposableTestActivity`. It seems that Google changed what is allowable in the manifest entries for Activities. The `ComposableTestActivity` is defined as part of the `androidTest` configuration, and so is not visible in the `debug` build flavor of the app. We need a solution that will make that Activity visible to the app.

This issue relates to:
- [ ] The Kotlin library
- [x] Compose Extension Library
- [ ] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**

1. Configure an emulator using API 33
2. Run the sample tests (`./gradlew sample:screenshotTest`)

Results:

```
android.content.ActivityNotFoundException: Unable to find explicit activity class {com.servicetitan.work.debug.test/androidx.test.core.app.InstrumentationActivityInvoker$BootstrapActivity}; have you declared this activity in your AndroidManifest.xml, or does your intent not match its declared <intent-filter>?
        at android.app.Instrumentation.checkStartActivityResult(Instrumentation.java:2158)
        at android.app.Instrumentation.execStartActivity(Instrumentation.java:1805)
        at androidx.test.runner.MonitoringInstrumentation.execStartActivity(MonitoringInstrumentation.java:603)
        at android.app.ContextImpl.startActivity(ContextImpl.java:1101)
        at android.content.ContextWrapper.startActivity(ContextWrapper.java:454)
        at androidx.test.core.app.InstrumentationActivityInvoker.startActivity(InstrumentationActivityInvoker.java:431)
        at androidx.test.core.app.InstrumentationActivityInvoker.startActivity(InstrumentationActivityInvoker.java:437)
        at androidx.test.core.app.ActivityScenario.launchInternal(ActivityScenario.java:265)
        at androidx.test.core.app.ActivityScenario.launch(ActivityScenario.java:195)
        at androidx.test.ext.junit.rules.ActivityScenarioRule.lambda$new$0$ActivityScenarioRule(ActivityScenarioRule.java:70)
        at androidx.test.ext.junit.rules.ActivityScenarioRule$$Lambda$0.get(ActivityScenarioRule.java:70)
        at androidx.test.ext.junit.rules.ActivityScenarioRule.before(ActivityScenarioRule.java:103)
[...]

PR #105, opened 2022-12-30
Issue-99: Upgrade sample to support API 33

DanielJette:

Resolves https://github.com/ndtp/android-testify/issues/99

- Upgrade Compose and Kotlin versions in Sample
    - Compose 1.0.5 -> 1.3.2
    - Kotlin 1.5.31 -> 1.7.20
- Add baseline for API 33

### DIFF
--- a/.github/workflows/sample_build.yml
+++ b/.github/workflows/sample_build.yml
@@ -11,6 +11,11 @@ jobs:
     - name: Fetch Sources
       uses: actions/checkout@v2
 
+    - name: Setup Java
+      uses: actions/setup-java@v1
+      with:
+        java-version: 1.8
+
     - name: Validate build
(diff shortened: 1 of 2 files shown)


realdm (approved):

lwise (approved):

DanielJette:
CI is breaking because (I think) it's using the wrong version of Java. I tried to fix it, but there are weird errors still:

```
exception: java.lang.UnsupportedClassVersionError: androidx/compose/compiler/plugins/kotlin/ComposeComponentRegistrar has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

I'll poke at this later and see if I can figure out what's wrong. In the meantime, I'm just going to let this sit here.

PR #106, opened 2023-01-01
Issue-98: ScreenshotTestRule should be backwards compatible

### LINKED ISSUE #98: [2.0] ScreenshotTestRule should be backwards compatible
Build on the concepts from https://github.com/ndtp/android-testify/compare/issue-60/java-builder (`ConfigurationBuilder`) and provide an API-compatible set of extensions for Testify 2.0 so that legacy code does not need to be migrated to the `configure {}` pattern immediately.

DanielJette:

Resolves #98 

### How have you achieved it?

This PR adds several extension methods for `ScreenshotRule` which are API-compatible with the 1.* Testify configuration methods. These extensions are intended to provide a backwards-compatible migration path for 2.0.0 upgrades. Instead of refactoring your tests to use the `configure` method, you can instead import the compatibility extensions to allow for an easier upgrade path.

This PR also adds a Java-compatible configuration builder and a new Edit Client test screen.

### Tophat instructions

I have added several new tests that use the legacy configuration format to test out the backwards compatibility.


lwise (approved):
two of the screenshots are named `BackwardsCompatibilityests`

Issue #97, opened 2022-09-30
[2.0] Configure should be generic


DanielJette:
The new `configure` method should return a generic.

Configure always returns a `ScreenshotTest<T>` but this means that subclasses of the Rule (like `ComposableScreenshotRule`) must be invoked prior to `configure`. It breaks the logical flow of the test where it's common to want to configure first, then set your composables. The configure method should be something more like:

```kotlin
fun <TRule : ScreenshotRule<T>> configure(configureRule: TestifyConfiguration.() -> Unit): TRule {
    configureRule.invoke(configuration)
    return this as TRule
}
```

PR #107, opened 2023-01-01
Issue-97: Open the `configure()` method so subclasses can return their own type

DanielJette:

Resolves https://github.com/ndtp/android-testify/issues/97

### How have you achieved it?

Open the `configure()` method so subclasses can return their own type

### DIFF
--- a/Sample/src/androidTest/java/dev/testify/sample/compose/ComposableScreenshotTest.kt
+++ b/Sample/src/androidTest/java/dev/testify/sample/compose/ComposableScreenshotTest.kt
@@ -157,15 +157,15 @@ class ComposableScreenshotTest {
         val locale = Locale.JAPAN
         val fontScale = 3.0f
         rule
+            .configure {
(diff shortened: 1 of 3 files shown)


Sample/src/androidTest/java/dev/testify/sample/compose/ComposableScreenshotTest.kt
```diff
@@ -157,16 +157,16 @@ class ComposableScreenshotTest {
         val locale = Locale.JAPAN
         val fontScale = 3.0f
         rule
+            .configure {
```
DanielJette, line 160: Without this change, you couldn't call `configure` here as it would return a rule instance cast as the base `ScreenshotRule` instead of the specific `ComposableScreenshotRule`. By returning `ComposableScreenshotRule`, you can now successfully chain other methods exposed by that class.

lwise (approved):

PR #113, opened 2023-03-11
Issue-41: Promote utility methods to top-level functions

DanielJette:

Relates to #41 

### How have you achieved it?

In order to promote extensibility, I have refactored `ScreenshotUtility` and `OutputFileUtility` to contain a collection of top-level functions, instead of being wrapped in singleton objects. These objects were holdover legacy from the Java version of Testify.

The functionality in these methods does not need to maintain any state nor hold any data. Top-level functions are simpler to write and easier to reason about than singleton objects. A top-level function can be used by any class, object or function and does not need to be tied to any specific class or object.

### Tophat instructions

No functionality should change. All tests should pass.


Ext/Fullscreen/src/main/java/dev/testify/capture/fullscreen/FullscreenCaptureMethod.kt
```diff
@@ -65,24 +69,23 @@ import java.io.File
...
     val instrumentation = InstrumentationRegistry.getInstrumentation()
     val fileName = getFileName(instrumentation.context)
-    val outputPath = OutputFileUtility().getOutputFilePath(activity, fileName)
+    val outputPath = getOutputFilePath(activity, fileName)
     val file = File(outputPath)
 
     // Use UiAutomator to take a screenshot
     val device = UiDevice.getInstance(instrumentation)
-    device.waitForIdle(1000)
+    device.waitForIdle(5000)
```
lwise, line 81: Check: does this need to be increased?

lwise (approved):

Ext/Fullscreen/src/main/java/dev/testify/capture/fullscreen/FullscreenCaptureMethod.kt
DanielJette: 🤔 In theory, it should be harmless ... it's the timeout on the wait. I was having some problems with the test failing; but it wasn't due to this timeout. I changed the timeout while debugging and forgot to put it back. I could probably revert it, but I think it's overall safer to have a slightly longer timeout here. So, I think I'd like to keep it.

PR #123, opened 2023-04-30
Throw an `UnexpectedDeviceException` if a baseline exists with a different device description than that of the currently running device.

DanielJette:

`ScreenshotRule.assertSame()` will now throw an `UnexpectedDeviceException` if a baseline exists with a different device description than that of the currently running device.
    This can be helpful to catch cases where you may be running a test against the wrong emulator. You can add add `-Pdevice=N` to any Testify Gradle command to target a specific device.

### Tophat instructions

Tests has been added


Library/src/androidTest/java/dev/testify/AssertExpectedDeviceTest.kt
```diff
@@ -0,0 +1,73 @@
...
+        rule.assertSame()
+    }
+
+    /**
+     * Ensure that missing a baseline is higher priority
+     */
+    @ScreenshotInstrumentation
+    @Test
+    fun testMissingBaseline() {
+        expectedException.expect(ScreenshotBaselineNotDefinedException::class.java)
```
lwise, line 56: These tests are kind of unobvious in how they work from the code; can we add a comment or something to specify which ones are expected to be run with a different-from-project-default emulator?

lwise (approved):

Library/src/androidTest/java/dev/testify/AssertExpectedDeviceTest.kt
DanielJette: Excellent point and great feedback. I'll definitely add some explanation in some comments

PR #124, opened 2023-05-01
Add documentation for Fullscreen Capture Method extension

DanielJette:

Add documentation for the Fullscreen Capture Method extension

### Tophat instructions

1. Checkout the branch `docs-fullscreen`
2. In the `/docs` folder, run `npm start`
3. Verify the documents in the _Extensions_ section

image

### DIFF
--- a/docs/docs/extensions/compose/_category_.json
+++ b/docs/docs/extensions/compose/_category_.json
@@ -1,6 +1,6 @@
 {
   "label": "Jetpack Compose",
-  "position": 1,
+  "position": 2,
   "link": {
     "type": "generated-index",
     "description": "Easily create screenshot tests for @Composable functions."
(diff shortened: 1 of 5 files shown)


docs/docs/extensions/compose/_category_.json
```diff
@@ -1,6 +1,6 @@
 {
   "label": "Jetpack Compose",
-  "position": 1,
+  "position": 2,
```
lwise, line 3: I guess these files are generated but why are both this and the fullscreen docs marked as position 2?

lwise (approved):

docs/docs/extensions/compose/_category_.json
DanielJette: It's because I made changes to add the Accessibility and Fullscreen docs at the same time, but put the PRs up separately since they technically don't depend on each other.
In the end, it's going to be ordered alphabetically so `0: Accessibility`, `1: Fullscreenshot`, `2: Jetpack Compose`

PR #129, opened 2023-05-05
Version bumps

andrew-carmichael:

Pretty much the idea is to tidy up the gradle scripts and update everything so the project will be able to accommodate a newer sample application. 

- modernize the gradle scripts 
- remove unneeded code from gradle scripts
- update gradle plugin and dependencies

Resolves <issues>

### How have you achieved it?
I did it in steps which are visible in the commits. 
1) bump android gradle plugin to 7.4.1 - this required me to change the Plugin project's `sourceCompatability` to Java 11 (this is either fine or very bad)
2) I relocated all of the repository blocks to be in settings.gradle
3) I updated anything that was not current version

### Tophat instructions
Passing CI should be sufficient. 

### Notice

### DIFF
--- a/Library/src/androidTest/java/dev/testify/RuleLifecycleTest.kt
+++ b/Library/src/androidTest/java/dev/testify/RuleLifecycleTest.kt
@@ -71,16 +71,17 @@ class RuleLifecycleTest {
         rule.assertSame()
     }
 
-    // Commenting this out for now.
-    // Testify throws MissingAssertSameException if a test case does not call assertSame. This test is proving that
-    // the exception is thrown. The problem is that ExpectedException is deprecated and I haven't been able to figure
-    // out how to make it work otherwise.
-//    @ScreenshotInstrumentation
-//    @Test
-//    fun testMethod3() {
-//        assertExpectedOrder(2, "testMethod3")
-//        assertExpectedOrder(3, "testMethod3")
-//    }
+    @Suppress("DEPRECATION")
+    @get:Rule val thrown: ExpectedException = ExpectedException.none()
+
+    @ScreenshotInstrumentation
+    @Test
+    fun testMethod3() {
+        assertExpectedOrder(2, "testMethod3")
+        assertExpectedOrder(3, "testMethod3")
+        thrown.expect(RuntimeException::class.java)
+        thrown.expectMessage("* You must call assertSame on the ScreenshotRule *")
+    }
 
     @UiThreadTest
     @ScreenshotInstrumentation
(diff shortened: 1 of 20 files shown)


build.gradle
```diff
@@ -52,26 +45,21 @@ buildscript {
                 'targetSdk' : 30
```
DanielJette, line 45: Looks like we need to bump the `compileSdk` to 33 for some of the dependencies. Did you give this a try?

andrew-carmichael: Yes but several other things go wrong after that.

Library/src/androidTest/java/dev/testify/RuleLifecycleTest.kt
```diff
@@ -74,15 +71,16 @@ class RuleLifecycleTest {
...
-        assertExpectedOrder(3, "testMethod3")
-
-        thrown.expect(RuntimeException::class.java)
-        thrown.expectMessage("* You must call assertSame on the ScreenshotRule *")
-    }
+    // Commenting this out for now.
+    // Testify throws MissingAssertSameException if a test case does not call assertSame. This test is proving that
+    // the exception is thrown. The problem is that ExpectedException is deprecated and I haven't been able to figure
+    // out how to make it work otherwise.
+//    @ScreenshotInstrumentation
```
DanielJette, line 78: This is a relatively important test that I'd like to keep.
It seems weird to me that they'd deprecate `ExpectedException` without providing a 1:1 replacement. There must be some way to do this. 🤔

Plugins/Gradle/build.gradle
```diff
@@ -31,8 +22,8 @@ version = "$project.versions.testify"
 group = pom.publishedGroupId
 archivesBaseName = pom.artifact
 
-
-sourceCompatibility = 1.8
+// Not able to upgrade gradle android plugin with this in place, Java 11 works, not sure if safe
```
DanielJette, line 25: Can remove the comment

Samples/Legacy/build.gradle
```diff
@@ -58,8 +50,7 @@ android {
         jvmTarget = "1.8"
     }
     composeOptions {
-        kotlinCompilerExtensionVersion versions.compose.core
-        kotlinCompilerVersion '1.7.21'
+        kotlinCompilerExtensionVersion "${versions.compose.compilerExt}"
```
DanielJette, line 53: I'd prefer if we uncoupled the Sample from the shared versions. Perhaps bring back the `versions` block and set `compileExt` to match whatever is in the root build.gradle?

DanielJette (changes requested):
This looks great. I'd just like to get that one test working again. I'll play with it locally and see if I can come up with an alternative.

Library/src/androidTest/java/dev/testify/RuleLifecycleTest.kt
andrew-carmichael: One would think!

DanielJette (approved):
Looks good to me

PR #149, opened 2023-06-21
Improve the use of Kotlin flows in the two use case classes improving the loading time of the two screens.

andrew-carmichael:
Enables the network requests to run in parallel drastically speeding up the load times for the main screen and the movie details screen.

Resolves <issues>

### How have you achieved it?
By wrapping the network requests in flows, and then using the combine function to collect the flows all at the same time.

### Tophat instructions
Launch the app and ensure the home screen as well as the movie details screen both load correctly.

### Notice


DanielJette (approved):
✅ Code
✅ Tested

Loads much faster!

andrew-carmichael:
Thank you!

Issue #153, opened 2023-07-05
Gradle plugin README mentions tasks which were removed


olegosipenko:
**Describe the bug**

This issue relates to:
- [ ] The Kotlin library
- [x] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [x] The documentation

**To Reproduce**
Imagemagick was removed in this PR https://github.com/ndtp/android-testify/pull/37 and also all the tasks to do manipulate with diffs and with ImageMagick. But Gradle plugin README still mentions those tasks.

**Expected behavior**
Documentation should be updated

PR #154, opened 2023-07-05
Removed diff and imagemagick tasks from README

olegosipenko:
Gradle Plugin README update

Resolves https://github.com/ndtp/android-testify/issues/153

### How have you achieved it?
...

### Tophat instructions
...

### Notice


DanielJette (approved):
Thank you for the contribution!

PR #156, opened 2023-07-12
Expose ScreenshotRule initializer value to ComposableScreenshotRule

philipsawyer:
Exposes the `enableReporter` initializer value from `ScreenshotRule` to the `ComposableScreenshotRule` so this value can be set when configuring a `ComposableScreenshotRule`.

Adds the ability to configure the `enableReporter` value from the `ComposableScreenshotRule` subclass. This initializer value was not accessible before and was hindering us from customizing the `ComposableScreenshotRule` as we wanted. This value is accessible via the superclass `ScreenshotRule` initializer. 

We found that setting `enableReporter` was more reliable than updating the `AndroidManifest` to enable the reporter (per Readme).

I also added documentation for this value based on what I saw around the repo but could use some help to make sure it is accurate. 

### How have you achieved it?
Added a new parameter to the `ComposableScreenshotRule` initializer and then passed this value to the superclass `ScreenshotRule`'s initializer. 

### Tophat instructions
...

### Notice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@
 #### Added
 
 - Reporter now supports skipped or ignored tests.
+- Added `enableReporter` parameter to `ComposableScreenshotRule` constructor
 
 ### Sample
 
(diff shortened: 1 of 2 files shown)


DanielJette:
Changes look great. Thank you for the contribution. Could you please include a brief entry to the [CHANGELOG.md](https://github.com/ndtp/android-testify/blob/main/CHANGELOG.md) for this PR?

philipsawyer:
Sure thing 👍  CHANGELOG.md has been updated now

DanielJette (approved):

PR #158, opened 2023-07-20
Create simple AsynchronousImage wrapper for Flix

andrew-carmichael:
Refactors Flix's use of Coil's AsyncImage Composable into a wrapper composable AsynchronousImage. The purpose of this is to isolate our use of AsyncImage such that it would be possible to move to a different image loader, or fix issues we run into. 

### How have you achieved it?
Simply created a wrapper composable.

### Tophat instructions
Ensure images in the application continue to be displayed as expected.

### Notice

### DIFF
--- a/Samples/Flix/src/main/java/dev/testify/samples/flix/ui/common/composeables/AsynchronousImage.kt
+++ b/Samples/Flix/src/main/java/dev/testify/samples/flix/ui/common/composeables/AsynchronousImage.kt
@@ -1,0 +1,49 @@
+package dev.testify.samples.flix.ui.common.composeables
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.ColorFilter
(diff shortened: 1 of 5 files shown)


Samples/Flix/src/main/java/dev/testify/samples/flix/ui/common/composeables/AsynchronousImage.kt
```diff
@@ -0,0 +1,49 @@
...
+    onError: ((AsyncImagePainter.State.Error) -> Unit)? = null,
+    alignment: Alignment = Alignment.Center,
+    contentScale: ContentScale = ContentScale.Fit,
+    alpha: Float = DefaultAlpha,
+    colorFilter: ColorFilter? = null,
+    filterQuality: FilterQuality = DrawScope.DefaultFilterQuality,
+) = AsyncImage(
+    model = model,
+    contentDescription = contentDescription,
+    imageLoader = LocalImageLoader.current,
```
DanielJette, line 36: Where's `LocalImageLoader` provided? Is that a free thing from coil?

DanielJette (approved):

PR #159, opened 2023-07-20
FlixSample, repair race condition between screenshot capture and Coil image loading

andrew-carmichael:
Fixes flaky tests that were caused by a race condition. Testify would capture its screenshots before the image loading library Coil had completed loading images. Coil loads images asynchronously so this is predicable. This is easily fixed by replacing Coil's coroutine dispatcher with one having Espresso IdlingResource awareness.

https://stackoverflow.com/questions/73044642/how-to-unit-test-if-an-image-was-loaded-using-coil-compose

### How have you achieved it?
I reconfigured Coil not to use its factory ImageLoader. I provide a custom ImageLoader that is configured to use the Espresso aware thread pool.

### Tophat instructions
- Ensure the screenshot tests pass for FlixSample
- Ensure the Flix sample app displays images as you would expect.

### Notice

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,8 @@
 
 ### Sample
 
+- FlixSample now using Espresso IdlingResource to eliminate screenshot test race condition.
+
 #### Changed
 
 - `Sample` has been renamed to `LegacySample` and moved to the `Samples/Legacy` directory.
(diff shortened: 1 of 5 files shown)


DanielJette (commented):
Could you add a CHANGELOG entry for this please?

DanielJette (approved):

PR #172, opened 2023-09-19
07 gradle managed device

### LINKED ISSUE #7: Add support for Gradle Managed Virtual Devices
**Is your feature request related to a problem? Please describe.**

Add support for Gradle Managed Virtual Devices

https://developer.android.com/studio/preview/features/#gmd

This feedback relates to:
- [x] The Kotlin library
- [x] The Gradle plugin
- [x] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

olegosipenko:

Add support of Gradle managed devices.
Resolves #7 

### How have you achieved it?

- Added new `useTestStorage` parameter to the `TestifyExtension` 
- Updated `TestifyPlugin` to support new parameter `useTestStorage`
- Added new parameter `isRecordMode` to the `ScreenshotRule` because with GMD we cannot run `screenshotRecord` task, but only able to run GMD-specific Gradle task
- Added new `TestStorageDestination`
- Updated `ScreenshotUtility.saveBitmapToDestination()` method to support new `Destination` type 
- Added new sample app
- Added new page to documentation

### Notice

> **Warning**


DanielJette:
Looks like the LegacySample needs to be updated:

> e: file:///bitrise/src/Library/src/test/java/dev/testify/ScreenshotRuleTest.kt:198:46 No value passed for parameter 'isRecordMode'

DanielJette:
We should run `./gradlew Plugin:assembleLocal` and add the modified plugin to this PR

docs/docs/recipes/20-gmd.md
```diff
@@ -0,0 +1,104 @@
...
+```
+
+### Running screenshot tests
+
+To perform screenshot test verification run any of the new Gradle tasks, for
+example `./gradlew testerDebugAndroidTest`.
+
+Because we need to run specific Gradle task to execute our tests using Gradle managed device, we
+cannot use `screenshotPull` task. When execution completed go to the
+module's `build/outputs/managed_device_android_test_additional_output/tester` folder to get recorded
```
DanielJette, line 74: This appears to be working pretty well, but I have noticed a couple of limitations that may require some refinement in follow-up PRs:

1. Testify supports various locales and screen orientations by encoding that information into the directory name (device key, eg. `30-1080x2220@440dp-en_US`). This information is lost in the GMD `managed_device_android_test_additional_output` directory. It would be beneficial to improve the file output for the Test Storage destination to preserve this information
2. Similar to above, it would be handy to preserve the device key as part of the `managed_device_android_test_additional_output` so that it would be easier to copy the screenshots to the proper `assets` directory
3. It would be really nice if the `screenshotPull` task were able to understand when a GMD device is in-use and correctly copy the files from `managed_device_android_test_additional_output` automatically. This should be possible by checking the state of the `useTestStorage` flag in `TestifySettings` -- similar to how we check the SDCard flag.

docs/docs/recipes/20-gmd.md
```diff
@@ -0,0 +1,104 @@
...
+
+```kotlin
+@get:Rule
+var rule = ScreenshotRule(ClientListActivity::class.java)
+
+@ScreenshotInstrumentation
+@Test
+fun testMissingBaseline() {
+    rule
+        .setRecordModeEnabled(true)
```
DanielJette, line 91: I think it would be preferable to set the record mode via a build.gradle flag, similar to `useTestStorage`.

e.g.

```
testify {
    isRecordMode System.getenv('TESTIFY_RECORD_MODE')
}
```
We already have an internal variable [`isRecordMode`](https://github.com/ndtp/android-testify/blob/2610898f25c8e42e5914858a6734bda6264b7ad5/Plugins/Gradle/src/main/kotlin/dev/testify/tasks/main/ScreenshotRecordTask.kt#L35C18-L35C30) being used by the `ScreenshotRecord` task which we can relocate to the `TestifySettings`. Then, we can configure that flag dynamically, without need to modify your code to record a baseline.

docs/docs/recipes/20-gmd.md
olegosipenko: 1-2. I am not sure about that, need to experiment
3. Yes, sounds like a good imrovement

docs/docs/recipes/20-gmd.md
olegosipenko: Then this would work globally, though when setting via rule we can set it just for some tests, not for all. WDYT?

docs/docs/recipes/20-gmd.md
olegosipenko: Regarding 1-2 I think the following: when running GMD, screenshots are stored in the `managed_device_android_test_additional_output` directory in the folder named after the name of the device given in the `managedDevices` clojure in the `build.gradle` file.
So, we have two options:
- either suggest user to name device following naming convention `VERSION-WIDTHxHEIGHT@DENSITY-LOCALE`
- or run some script after gradle task to rename the folder ?

olegosipenko: For 3 point we need to read `managedDevices` clojure in the `build.gradle` to understand where take screenshots from. I think it's also better to keep for further improvement

docs/docs/recipes/20-gmd.md
DanielJette: Yes, that's a good point about being global. I think it comes down to what kind workflow the developer would have when writing tests.
When I made my comment, I was thinking about running this on a CI service -- that you'd want to set that flag via an environment variable that would control "record mode" globally so that you could update your screenshots using a CI service.

But, your question brings up a very valid flaw in my logic. What I think that we need is a convenient way to record a screenshot without needing to modify the code. I can't think of such a way right now, so I believe now that your current solution is acceptable and the best we can do for right now.

I think some of this could be addressed by updating both `screenshotTest` and `screenshotRecord` to work with the GMD gradle tasks so that they could configure Testify to work as expected. Also, there's some potential to update the IntelliJ Platform Plugin to support the GMD tasks too.

In any case, let's go with what you have now and we can always improve or enhance this later.

DanielJette: Related to the above, can you describe your typical workflow when using GMD? As I've never used them myself, I'm curious how they fit in your development cycle.

olegosipenko: We use them to run screenshot tests on CI - it's easier to write few lines in `build.gradle`, than managing emulator. But when we need to update existing or create new baseline we do it locally

olegosipenko: I think I will add this parameter to `TestifyExtension` and will keep the way to enable record mode in the rule too

DanielJette: Good idea

docs/docs/recipes/20-gmd.md
DanielJette: Yes, these are good ideas for future improvement -- don't need to address this now

DanielJette (approved):
Amazing work on this. Thank you for the contribution!

PR #183, opened 2023-10-23
Add Jacoco for code coverage and add unit tests

DanielJette:

This PR was created in anticipation of the final release of Testify 2.0. It was my desire to ensure that all core code paths were tested and verified. This increases the confidence in the release and protects against regressions. One of the primary goals of the Testify 2.0 architecture is to be modular and extensible. With these tests in place, it ensures that future extensions to Testify will maintain backward compatibility and the API contract established in the 2.0 release.

I have integrated [Jacoco](https://github.com/jacoco/jacoco) to calculate code coverage. My goal was a minimum of 80% coverage on every package, and an overall minimum coverage of 90%.

After this PR, the coverage is at 93% for the entire Library. Only 465 of 6,988 instructions remain uncovered.

### How have you achieved it?

- Remove unreachable and unused code
- Make some methods or classes more testable
- Add more tests

Bitrise CI verification should be sufficient.

image

### Notice

> **Warning**


Library/src/main/java/dev/testify/core/processor/diff/HighContrastDiff.kt
```diff
@@ -47,13 +47,18 @@ import dev.testify.saveBitmapToDestination
...
 
-    private lateinit var fileName: String
-    private lateinit var baselineBitmap: Bitmap
-    private lateinit var currentBitmap: Bitmap
+    private var fileName: String? = null
+    private var baselineBitmap: Bitmap? = null
+    private var currentBitmap: Bitmap? = null
 
     fun generate(context: Context) {
+        val fileName = this.fileName ?: throw IllegalArgumentException("call name() to set fileName")
```
lwise, line 57: 👌

lwise (approved):

PR #222, opened 2024-06-02
215: Upgrade libraries to version catalog

DanielJette:

Update the libraries to use a version catalog (`libs.versions.toml`)

- Library
- ComposeExtensions
- Accessibility
- FullscreenCaptureMethod
- Plugin

Note that the Samples are not using the version catalog. This is because the samples intentionally are not required to use the same versions as the core library projects.


jaredh (approved):
LGTM.

AndroidTestifyBot (approved):

PR #229, opened 2024-06-29
Use hardware accelerated GPU on Bitrise emulator

DanielJette:

The Ubuntu 22 stack from Bitrise supports [kvm hardware acceleration](https://developer.android.com/studio/run/emulator-acceleration) meaning we can use real GPU rendering, not software rendering. 
This will improve the performance of the emulator.


DanielJette:
Makes no difference. Not needed

Issue #233, opened 2024-09-02
Plugin not compatible with latest Canary version of Android Studio


Dukoff92:
Hello,

Just a heads up that version 2.3.1 of Android Testify plugin is not compatible with the latest Android Studio Ladybug 2024.2.1 Canary 9 (Build #AI-242.20224.300.2421.12293307).

DanielJette:
Thanks for the bug report @Dukoff92 
An updated was submitted on 2024-09-02 and is pending approval by Jetbrains

Dukoff92:
Thanks for the swift reaction!

DanielJette:
@Dukoff92 2.4 should be available from the Intellij Marketplace now

PR #234, opened 2024-09-03
233: Added support for Android Studio Ladybug | 2024.2.1 Canary 9 | 242.+ to IntelliJ plugin

DanielJette:

Resolves https://github.com/ndtp/android-testify/issues/233

Release 2.4.0 with added support for Android Studio Ladybug | 2024.2.1 Canary 9 | 242.+

### Notice

> [!WARNING]

### DIFF
--- a/.github/workflows/central_release.yml
+++ b/.github/workflows/central_release.yml
@@ -1,7 +1,5 @@
 name: Publish package to the Maven Central Repository
 on:
-  release:
-    types: [created]
   workflow_dispatch:
 jobs:
   publish:
--- a/Plugins/IntelliJ/CHANGELOG.md
+++ b/Plugins/IntelliJ/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Android Testify - IntelliJ Platform Plugin - Change Log
 
+## [2.4.0]
+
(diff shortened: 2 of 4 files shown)


Plugins/IntelliJ/build.gradle.kts
```diff
@@ -60,6 +60,6 @@ tasks {
     runIde {
         // Absolute path to installed target 3.5 Android Studio to use as
         // IDE Development Instance (the "Contents" directory is macOS specific):
-        ideDir.set(file("/Applications/Android Studio.app/Contents"))
+        ideDir.set(file("/Applications/Android Studio Ladybug | 2024.2.1 Canary 9.app/Contents"))
```
lwise, line 63: Is this from having multiple Android Studio versions installed or have they changed the default install naming?

DanielJette: This is an oversight on my part. Didn't mean to check out that change. The default name for Android Studio hasn't changed, I was just testing this with an install of Canary 9 that I had installed side-by-side with my normal Android Studio.

I'll revert this change

AndroidTestifyBot (approved):

PR #235, opened 2024-09-15
203: Support NO_COLOR

### LINKED ISSUE #203: Support NO_COLOR
An increasing number of command-line software programs output text with [ANSI color](https://en.wikipedia.org/wiki/ANSI_escape_code#Colors) escape codes by default. While some developers and users obviously prefer seeing these colors, some users don’t. Unfortunately, every new piece of software seems to have a [different way](https://no-color.org/#other) of disabling colored text output and some software has no way at all.

Accepting the futility of trying to reverse this trend, an informal standard was proposed in 2017:

Command-line software which adds ANSI color to its output by default should check for a NO_COLOR environment variable that, when present and not an empty string (regardless of its value), prevents the addition of ANSI color.

By adopting this standard, users that prefer to have plain, non-colored text output can export NO_COLOR=1 to their shell’s environment and automatically disable color by default in all supported software.

If your software outputs color by default, please consider not doing so. If you insist, please implement this standard to make it easy for your users to disable color and then add your software to this list by [submitting a pull request](https://github.com/jcs/no_color).

See: https://no-color.org/

DanielJette:

Resolves #203 

### How have you achieved it?

Add support for `NO_COLOR`. See https://no-color.org/ for more information.

Set `export NO_COLOR=true` and run various testify command (ex. `./gradlew LegacySample:testifySettings`) and verify that no color is used in the output.

### Notice

> [!WARNING]

### DIFF
--- a/Plugins/Gradle/src/main/kotlin/dev/testify/tasks/utility/SettingsTask.kt
+++ b/Plugins/Gradle/src/main/kotlin/dev/testify/tasks/utility/SettingsTask.kt
@@ -26,7 +26,10 @@
 package dev.testify.tasks.utility
 
 import dev.testify.internal.Adb
+import dev.testify.internal.AnsiFormat
 import dev.testify.internal.Device
+import dev.testify.internal.ENV_NO_COLOR
+import dev.testify.internal.isEnvSet
 import dev.testify.internal.reportFilePath
 import dev.testify.internal.screenshotDirectory
 import dev.testify.tasks.internal.TaskNameProvider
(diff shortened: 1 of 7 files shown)


lwise (approved):

Plugins/Gradle/src/main/kotlin/dev/testify/tasks/utility/SettingsTask.kt
```diff
@@ -96,8 +134,11 @@ open class SettingsTask : TestifyUtilityTask() {
         println("  testRunner             = $testRunner")
         println("  useSdCard              = $useSdCard")
         println("  useTestStorage         = $useTestStorage")
-        println("  isRecordMode           = $isRecordMode")
         println("  user                   = $userId")
+
+        val notSet = "${AnsiFormat.Bold}${AnsiFormat.Green}Not set${AnsiFormat.Reset}"
```
jaredh, line 139: Nice touch.

jaredh (approved):
TIL NO_COLOR.

✅ Tested

Might be useful to respect Gradle's console output mode for this as well.

DanielJette:
> Might be useful to respect Gradle's console output mode for this as well.

@jaredh I haven't heard of Gradle's console output mode before. Do you mean this? https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_customizing_log_format

jaredh:
> > Might be useful to respect Gradle's console output mode for this as well.
> 
> @jaredh I haven't heard of Gradle's console output mode before. Do you mean this? [docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_customizing_log_format](https://docs.gradle.org/current/userguide/command_line_interface.html#sec:command_line_customizing_log_format)

@DanielJette Yep, thats the one. `plain` should have no colour.

DanielJette:
https://github.com/ndtp/android-testify/pull/239 is a better solution

PR #239, opened 2024-11-16
203: Use Gradle's StyledTextOutput for log formatting

jaredh:

Resolves https://github.com/ndtp/android-testify/issues/203

Alternative take on https://github.com/ndtp/android-testify/pull/235

### How have you achieved it?

Uses Gradle's StyledTextOutput to format log output, allowing us to respect console mode and console color configuration.

Since Testify is a plugin, I think it makes more sense to integrate with Gradle itself than `NO_COLOR`.

Supported styles are:

Screenshot 2024-11-16 at 4 28 38 PM

1. `./gradlew FlixLibrary:testifySettings -Pverbose=true` shows coloured output
2. `./gradlew FlixLibrary:testifySettings -Pverbose=true --console=plain` shows no coloured output
3. ` ./gradlew -Dorg.gradle.color.description=BLUE FlixLibrary:testifySettings -Pverbose=true --console=plain` shows customizing colour output

### Notice

> [!WARNING]

### DIFF
--- a/Plugins/Gradle/src/main/kotlin/dev/testify/TestifyPlugin.kt
+++ b/Plugins/Gradle/src/main/kotlin/dev/testify/TestifyPlugin.kt
@@ -27,7 +27,7 @@ package dev.testify
 
 import dev.testify.TestifyPlugin.Companion.EVALUATED_SETTINGS
 import dev.testify.internal.Adb
-import dev.testify.internal.AnsiFormat
+import dev.testify.internal.Style.Description
 import dev.testify.internal.android
 import dev.testify.internal.isVerbose
 import dev.testify.internal.println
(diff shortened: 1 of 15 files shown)


Plugins/Gradle/src/main/kotlin/dev/testify/TestifyPlugin.kt
```diff
@@ -127,6 +129,7 @@ class TestifyPlugin : Plugin<Project> {
     }
 
     companion object {
+        var consoleColorEnabled = true
```
jaredh, line 132: @DanielJette I don't like this at all, but didn't like passing around project either. Have any suggestions or alternatives? `TestifySettings`?

DanielJette: I looked into this shortly after you originally suggested using the console mode and came to the same conclusion. With the architecture of the gradle plugin, I don't think there's another choice.

This is consistent with other solutions we're using in the plugin, at least.

Plugins/Gradle/src/main/kotlin/dev/testify/tasks/utility/SettingsTask.kt
```diff
@@ -81,10 +117,16 @@ open class SettingsTask : TestifyUtilityTask() {
 
     override fun taskAction() {
         val userId = Adb.forcedUser?.toString() ?: Device.user.takeUnless { Device.isEmpty } ?: "Device not found"
+        val notSet =
+            "${AnsiFormat.Bold}${AnsiFormat.Green}${project.gradle.startParameter.consoleOutput}${AnsiFormat.Reset}"
```
DanielJette, line 121: Using `project` in here breaks the gradle configuration cache

jaredh: Yeah, I think it makes sense to just not expose this in settings at all since it isn't unique to the plugin.

PR #247, opened 2025-02-16
Revert "203: Use Gradle's StyledTextOutput for log formatting"

DanielJette:
Reverts ndtp/android-testify#239

Unfortunately gradle 8.10.* has moved `StyledTextOutput` to an internal package and it can not longer be referenced.

It seems like it was always intended to be an internal API: https://discuss.gradle.org/t/bug-in-gradle-2-14-rc1-no-service-of-type-styledtextoutputfactory/17638

```
e: file:///Users/danjette/dev/android-testify/Plugins/Gradle/src/main/kotlin/dev/testify/TestifyPlugin.kt:54:49 Unresolved reference: serviceOf
e: file:///Users/danjette/dev/android-testify/Plugins/Gradle/src/main/kotlin/dev/testify/TestifyPlugin.kt:62:40 Unresolved reference: serviceOf
```

@jaredh do you know of any other ways to access this? I suppose we could try reflection, but that feels wrong


jaredh (changes requested):
> It seems like it was always intended to be an internal API: [discuss.gradle.org/t/bug-in-gradle-2-14-rc1-no-service-of-type-styledtextoutputfactory/17638](https://discuss.gradle.org/t/bug-in-gradle-2-14-rc1-no-service-of-type-styledtextoutputfactory/17638)

Took a look at this and had no problems referencing `StyledTextOutput` on Gradle `8.10.2`. `serviceOf` however seems to have moved and is available now via 👇 

```kotlin
import org.gradle.internal.extensions.core.serviceOf
```

Yes it's internal, but tacitly public and IMO still the most appropriate way to provide stylized text through a Gradle plugin. There's a long running Gradle issue for exposing this functionality officially, worth a ⭐ https://github.com/gradle/gradle/issues/5825.

I think this can be closed in favour of updating the `serviceOf` import.

DanielJette:
Thanks @jaredh. I'll fix it up that way

PR #248, opened 2025-02-17
244: Fix GMD recordMode not working as expected

### LINKED ISSUE #244: GMD recordMode not working as expected
Setting up GMD, followed the instructions in both the blog and the recipe section but can't seem to get the second way of recording baseline to work. By putting `recordMode true` in the `build.gradle` file instead of through the test configure method.

This issue relates to:
- [ ] The Kotlin library
- [x] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**
Try setting the recordMode to true and record baseline

**Expected behavior**
New baseline are recorded and do not get failures in build with the `dev.testify.core.exception.ScreenshotIsDifferentException:`

**Desktop (please complete the following information):**
 - OS: macOS
 - Version 15.3

**Target Android Device (please complete the following information):**

```
 testOptions {
        managedDevices {
            devices {
                  screenshotTest(ManagedVirtualDevice) {
                      device = "Pixel 4"
                      apiLevel = 34
                      systemImageSource = "aosp"
                 }
            }
        }
    }
```

**Additional context**
```
    testify {
        applicationPackageId "..."
        testPackageId "..."
        useTestStorage true
        recordMode true
    }
```

### LINKED ISSUE #246: Upgrade to Gradle 8.10.2

DanielJette:

- Fixes #244
- Fixes #246 
- Reverts #235

### How have you achieved it?

Record mode in GMD tests were not working since the `recordMode` flag was not being passed from the `testify {}` closure to the testify library.

The version of [`getApplicationInfo()`](https://developer.android.com/reference/android/content/pm/PackageManager#getApplicationInfo(java.lang.String,%20android.content.pm.PackageManager.ApplicationInfoFlags)) added in API 33 was not returning the expected manifest data. Reverting to the [original](https://developer.android.com/reference/android/content/pm/PackageManager#getApplicationInfo(java.lang.String,%20int)) (since API 1) version of `getApplicationInfo` works as expected.

@matthew-shin-hs

### Notice

> [!WARNING]


lwise (approved):

DanielJette:
> It looks like we can't build the `Plugin` module with JDK 21 because of mismatched default versions.
> 
> Could we drive-by address this here since we're bumping Gradle in this PR? Something like this in `Plugin/Gradle/build.gradle` should do it...
> 
> ```groovy-gradle
> java {
>     toolchain {
>         languageVersion = JavaLanguageVersion.of(17)
>     }
> }
> ```

Should I just update Bitrise to use java 21?

jaredh:
> Should I just update Bitrise to use java 21?

That will still require migrating the Kotlin tasks to use `21` - currently they're set to `17`. Off the cuff, not sure if there's any downstream impact to targeting a higher JVM but there's no harm continuing to target `17`. The problem is some targets are currently _unspecified_ so they default to using the current JDK version.

jaredh (approved):

PR #303, opened 2026-08-11
Add sample app demonstrating Testify usage with Paparazzi

DanielJette:

Adds a sample app, based on the existing Flix sample, to demonstrate testing using [Paparazzi](https://github.com/cashapp/paparazzi).

### How have you achieved it?

The existing Flix sample was copied and migrated to using Paparazzi. A basic PaparazziTestRule was added to the sample app. The intention is for this to be a jumping-off point for using Testify alongside Paparazzi. Follow-up PRs will include Paparazzi extensions and migration methods, allowing you to chose which rendering/capture pipeline to use with Testify. Support for Paparazzi and classic Testify will be added to the gradle and Intellij plugins allowing you to keep the efficiency and ease-of-use of Testify, while giving you the flexibility to chose between emulator or preview-based screenshot testing.

### Notice

> [!WARNING]


bitrise.yml
```diff
@@ -430,6 +431,44 @@ workflows:
...
+        - project_location: "$PROJECT_LOCATION"
+        - variant: Debug
+        - module: PaparazziSample
+        - cache_level: none
+    - android-unit-test@1:
+        inputs:
+        - project_location: "$PROJECT_LOCATION"
+        - variant: Debug
+        - module: PaparazziSampleLibrary
+        - cache_level: none
```
AndroidTestifyBot: **This never verifies the snapshots.**

`android-unit-test@1` runs `testDebugUnitTest`. Paparazzi only compares against the recorded goldens when `verifyPaparazzi<Variant>` is in the task graph — `PaparazziPlugin` derives `paparazzi.test.verify` from `graph.hasTask(verifyTaskProvider)`, and `Paparazzi.determineHandler` falls back to `HtmlReportWriter` (which asserts nothing) when that property is false.

I confirmed it by replacing one golden with a different image:

```
$ ./gradlew :PaparazziSampleLibrary:testDebugUnitTest --rerun-tasks
BUILD SUCCESSFUL

$ ./gradlew :PaparazziSampleLibrary:verifyPaparazziDebug --rerun-tasks
CastDetailScreenshotTest > loading FAILED
BUILD FAILED
```

Both checks go green regardless of what the UI actually renders. `Samples/Paparazzi/README.md` already documents the right command — CI should use it:

```yaml
    - gradle-runner@2:
        inputs:
        - gradlew_path: "./gradlew"
        - gradle_task: PaparazziSample:verifyPaparazziDebug
        title: Verify Paparazzi Snapshots
    - gradle-runner@2:
        inputs:
        - gradlew_path: "./gradlew"
        - gradle_task: PaparazziSampleLibrary:verifyPaparazziDebug
        title: Verify Paparazzi Snapshots (Library)
```

bitrise.yml
```diff
@@ -430,6 +431,44 @@ workflows:
...
+        inputs:
+        - auth_token: $GITHUB_TOKEN
+        - status_identifier: "Paparazzi Sample App"
+        - pipeline_build_url: "$BITRISE_BUILD_URL"
+    - custom-test-results-export@1.0:
+        is_always_run: true
+        timeout: 120
+        inputs:
+        - test_name: Paparazzi Screenshot Test
+        - search_pattern: "**/build/test-results/"
```
AndroidTestifyBot: `"**/build/test-results/"` is a directory, not a file glob — every other workflow in this file matches files (`"**test_result_**.xml**"`). The reports land at `*/build/test-results/testDebugUnitTest/TEST-*.xml`, so something like `"**/build/test-results/**/*.xml"` is probably what's wanted. Worth checking the step actually exports anything today.

bitrise.yml
```diff
@@ -430,6 +431,44 @@ workflows:
...
+        - notify_user_groups: none
+    - github-status@3:
+        run_if: .IsCI
+        inputs:
+        - auth_token: $GITHUB_TOKEN
+        - status_identifier: "Paparazzi Sample Library"
+        - pipeline_build_url: "$BITRISE_BUILD_URL"
+    before_run:
+    - _globalSetup
+    - _emulatorSetup
```
AndroidTestifyBot: Paparazzi renders on the JVM with layoutlib — no device or emulator involved. `_emulatorSetup` boots an AVD (240s launch timeout + a 300s wait step) that nothing in this workflow uses. Dropping it should leave just `before_run: [_globalSetup]`.

(Related, and outside the diff so I can't comment inline: `evaluate_build` sets a `"pending"` status for all nine existing identifiers but not for "Paparazzi Sample App" / "Paparazzi Sample Library".)

Samples/Paparazzi/PaparazziSampleLibrary/build.gradle
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+    buildFeatures {
+        compose true
+    }
+    // Publishes the `src/testFixtures` source set as a consumable variant so that :PaparazziSample
+    // (and any other module migrating from Testify) can share PaparazziTestRule.
+    testFixtures {
+        enable true
+    }
```
AndroidTestifyBot, line 61: **Missing `forkEvery = 1`.**

The app module sets this with a long comment explaining that without one JVM per test class, every class after the first records a *blank* image while still passing. This module has exactly the same exposure — `CastDetailScreenshotTest.loaded` pulls `Newton.png` through Coil — and no `testOptions` block. It's only safe today because there is a single Paparazzi test class here; adding a second one silently reintroduces the bug the app module documents.

```groovy
    testOptions {
        unitTests.all {
            // See :PaparazziSample — one JVM per Paparazzi test class.
            forkEvery = 1
        }
    }
```

Samples/Paparazzi/PaparazziSampleLibrary/README.md
```diff
@@ -0,0 +1,36 @@
+# Paparazzi
+
+A sample demonstrating the interoperability of Testify with Paparazzi for a library project.
+
+
+## Building
+
+```
+./gradlew PaparazziLibrarySample:assembleDebug
```
AndroidTestifyBot: Module is `PaparazziSampleLibrary`, so this should be `./gradlew PaparazziSampleLibrary:assembleDebug` — `PaparazziLibrarySample` doesn't exist.

Also worth adding the Paparazzi commands here (this module has its own goldens under `src/test/snapshots/images/`), and the copyright below says 2025 while the rest of the PR says 2026.

Samples/Paparazzi/README.md
```diff
@@ -0,0 +1,77 @@
...
+        }
+    }
+}
+```
+
+Image fixtures are checked in under `src/debug/assets/images/` and referenced with the same
+`file:///android_asset/...` URIs an instrumented test would use — Paparazzi resolves assets against
+the module's merged asset directories, so Coil loads them exactly as it does on a device.
+
+`PaparazziSampleLibrary` keeps one Testify test, so the repository still exercises Testify against a
```
AndroidTestifyBot: This claim doesn't hold as wired: `PaparazziSampleLibrary` doesn't apply the `dev.testify` plugin (unlike `FlixLibrary`), so there's no `screenshotTest` task, and `test_paparazzi` only runs the JVM unit tests — nothing in CI ever executes the instrumented test or its baselines under `src/androidTest/assets/screenshots/29-1080x2220@440dp-en_US/`.

Either apply the plugin and add a `screenshotTest` step to the workflow, or reword to say the test is kept as a reference rather than exercised.

gradle/libs.versions.toml
```diff
@@ -76,4 +78,5 @@ google-dagger-hilt-android = { id = "com.google.dagger.hilt.android", version.re
 jetbrains-dokka = { id = "org.jetbrains.dokka", version.ref = "dokka" }
 jetbrains-kotlin-android = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
 jetbrains-kotlin-plugin-serialization = { id = "org.jetbrains.kotlin.plugin.serialization", version.ref = "kotlinPluginSerialization" }
-ktlint = { id = "org.jlleitschuh.gradle.ktlint", version.ref = "ktlint"}
+ktlint = { id = "org.jlleitschuh.gradle.ktlint", version.ref = "ktlint" }
+paparazzi = { id = "app.cash.paparazzi", version.ref = "paparazzi" }
```
AndroidTestifyBot: Trailing newline got dropped here (`\ No newline at end of file`). `.editorconfig` sets `insert_final_newline=true`, and the file had one before this change.

Samples/Paparazzi/src/main/java/dev/testify/samples/paparazzi/application/PaparazziApplication.kt
```diff
@@ -0,0 +1,32 @@
...
+ * THE SOFTWARE.
+ */
+
+package dev.testify.samples.paparazzi.application
+
+import android.app.Application
+import dagger.hilt.android.HiltAndroidApp
+
+@HiltAndroidApp
+class PaparazziApplication : Application()
```
AndroidTestifyBot, line 32: Class is `PaparazziApplication` but the file is still `FlixApplication.kt` — ktlint's `filename` rule flags this (`The file should be named after the class, 'PaparazziApplication.kt'`). The library module did rename `FlixPerson` → `PaparazziPerson`, so this looks like a straggler.

AndroidTestifyBot (changes requested):
## Paparazzi sample — review

The migration itself is solid. The main sources are a faithful copy of the Flix sample (verified: identical modulo the `flix` → `paparazzi` package rename, the `composeables` → `composables` typo fix, and added license headers), so the real review surface is the build/CI wiring and the new `testFixtures` support code. `PaparazziTestRule` and `PaparazziImageLoader` are genuinely good — the `Dispatchers.Main` / dead-`Looper` explanation and the `forkEvery = 1` rationale are the kind of comments that save the next person a day.

**Verified locally:** `:PaparazziSample:verifyPaparazziDebug` and `:PaparazziSampleLibrary:verifyPaparazziDebug` both pass against the checked-in goldens. The new test and `testFixtures` sources are ktlint-clean (the copied `src/main` sources are not, but neither are Flix's, and no CI workflow runs `ktlintCheck` on samples — consistent with existing practice).

Two blocking issues, both in the "silently green" category this PR is otherwise careful about.

---

### 🔴 1. CI never actually verifies the snapshots

`test_paparazzi` runs the tests via `android-unit-test@1`, which executes `testDebugUnitTest`. Paparazzi only compares against goldens when `verifyPaparazzi<Variant>` is in the task graph:

```kotlin
// PaparazziPlugin.kt
isVerifyRun.set(verifyTaskProvider.map { graph.hasTask(it) })
test.systemProperties["paparazzi.test.verify"] = isVerifyRun.get()

// Paparazzi.kt
private fun determineHandler(maxPercentDifference: Double): SnapshotHandler =
  if (isVerifying) SnapshotVerifier(...) else HtmlReportWriter(...)
```

Without the flag it writes an HTML report and asserts nothing. Confirmed empirically — I swapped one golden for a different image and re-ran both:

| command | result |
|---|---|
| `:PaparazziSampleLibrary:testDebugUnitTest` (what CI runs) | **BUILD SUCCESSFUL** |
| `:PaparazziSampleLibrary:verifyPaparazziDebug` | `CastDetailScreenshotTest > loading FAILED` |

So as written, the "Paparazzi Sample App" / "Paparazzi Sample Library" checks go green no matter what the UI looks like. `Samples/Paparazzi/README.md` already documents the right command; CI just needs to use it. Swapping the two `android-unit-test@1` steps for `gradle-runner@2` with `gradle_task: PaparazziSample:verifyPaparazziDebug` (and the library equivalent) fixes it.

### 🔴 2. `forkEvery = 1` is missing from `:PaparazziSampleLibrary`

The app module explains at length that without one JVM per test class, every class after the first records blank images *while still passing*. The library module has the same exposure — `CastDetailScreenshotTest.loaded` loads `Newton.png` through Coil — but no `testOptions` block. It works today only because that module happens to have exactly one Paparazzi test class. The moment a second one is added, the failure mode is the silent one the app module's comment warns about.

---

### Non-blocking

- **`_emulatorSetup` in `test_paparazzi`** — Paparazzi is JVM-only. This boots an AVD (240s launch timeout + 300s wait) that nothing uses.
- **No pending status for the new checks** — `evaluate_build` sets `"pending"` for all nine existing status identifiers but not for "Paparazzi Sample App" / "Paparazzi Sample Library" (bitrise.yml:33-95, outside the diff so I couldn't comment inline). If the workflow doesn't start, those checks are absent rather than pending.
- **`custom-test-results-export` search pattern** — `"**/build/test-results/"` is a directory; every other workflow in the file uses a file glob (`"**test_result_**.xml**"`). The XMLs live at `*/build/test-results/testDebugUnitTest/TEST-*.xml`. Worth confirming it matches anything.
- **The library's Testify test is never run.** `PaparazziSampleLibrary` ships a `ComposableScreenshotRule` test plus recorded baselines under `src/androidTest/assets/screenshots/29-1080x2220@440dp-en_US/`, and the README says the module "keeps one Testify test, so the repository still exercises Testify against a library project." But the module doesn't apply the `dev.testify` plugin (so there's no `screenshotTest` task, unlike `FlixLibrary`) and CI only runs its JVM unit tests. Either wire it up or soften the README claim.

### Nits

- `Samples/Paparazzi/src/debug/assets/images/headshots/Newton.png` (518 KB) isn't referenced by any test in the app module — only the library's copy is used. Inherited from Flix, but easy to drop from a fresh copy.
- `flix.svg` / `logo_flix.svg` at the sample root don't appear to be referenced from anything in this sample.

Happy to re-review once the CI wiring is sorted — the code itself is in good shape.

bitrise.yml
```diff
@@ -430,6 +438,41 @@ workflows:
...
+        inputs:
+        - auth_token: $GITHUB_TOKEN
+        - status_identifier: "Paparazzi Sample App"
+        - pipeline_build_url: "$BITRISE_BUILD_URL"
+    - custom-test-results-export@1.0:
+        is_always_run: true
+        timeout: 120
+        inputs:
+        - test_name: Paparazzi Screenshot Test
+        - search_pattern: "**/build/test-results/"
```
AndroidTestifyBot: Still unchanged from last round — `"**/build/test-results/"` is a directory, so this step has nothing to export. I confirmed the actual layout on a failing run:

```
Samples/Paparazzi/PaparazziSampleLibrary/build/test-results/testDebugUnitTest/TEST-dev.testify.samples.paparazzi.ui.cast.CastDetailScreenshotTest.xml
```

so `"**/build/test-results/**/*.xml"` would match.

While you're here: now that verification is real, the artifact worth keeping on a red build is Paparazzi's visual diff, which lands in `build/paparazzi/failures/debug/` as `<name>.png` plus `delta-<name>.png` (verified). Nothing currently collects it — a `deploy-to-bitrise-io` `deploy_path` or a copy into `$BITRISE_DEPLOY_DIR` would make failures diagnosable from the Bitrise UI instead of just "1 test failed". Happy for this to be a follow-up.

bitrise.yml
```diff
@@ -92,6 +93,13 @@ workflows:
         - auth_token: $GITHUB_TOKEN
         - status_identifier: "Fullscreen Extension"
         - pipeline_build_url: "$BITRISE_BUILD_URL"
+    - github-status@3:
+        run_if: .IsCI
+        inputs:
+        - set_specific_status: "pending"
+        - auth_token: $GITHUB_TOKEN
+        - status_identifier: "Paparazzi Sample App"
+        - pipeline_build_url: "$BITRISE_BUILD_URL"
```
AndroidTestifyBot, line 102: Only "Paparazzi Sample App" got a pending status — "Paparazzi Sample Library" is still reported at the end of `test_paparazzi` (line 472) but never marked pending here, so it's the one identifier out of eleven without one.

```yaml
    - github-status@3:
        run_if: .IsCI
        inputs:
        - set_specific_status: "pending"
        - auth_token: $GITHUB_TOKEN
        - status_identifier: "Paparazzi Sample Library"
        - pipeline_build_url: "$BITRISE_BUILD_URL"
```

AndroidTestifyBot (approved):
## Re-review — 83a6d9e

Both blocking issues are fixed, and I verified the CI fix end-to-end rather than just reading it.

### Blockers resolved

**1. CI now actually verifies.** The `gradle-runner@2` / `verifyPaparazziDebug` swap works. Repeating the same experiment as last time — replacing one golden with a different image:

| | before (`testDebugUnitTest`) | now (`verifyPaparazziDebug`) |
|---|---|---|
| bad golden | BUILD SUCCESSFUL ❌ | `CastDetailScreenshotTest > loading FAILED` ✅ |

And with the goldens as committed, `:PaparazziSample:verifyPaparazziDebug` + `:PaparazziSampleLibrary:verifyPaparazziDebug` both pass from a forced re-run.

**2. `forkEvery = 1`** is in `PaparazziSampleLibrary`, pointing back at the app module's explanation. Good.

Also confirmed fixed: `_emulatorSetup` dropped from `test_paparazzi` (saves the AVD boot + 300s wait), `libs.versions.toml` trailing newline restored, `FlixApplication.kt` → `PaparazziApplication.kt`, library README module name and copyright year, and the app README no longer claims CI exercises the Testify test.

### Two small things left over — non-blocking

Both are inline. Neither affects whether a regression gets caught, only how legible it is when one does, so I'm not holding the PR on them:

- The `custom-test-results-export` pattern is unchanged and still doesn't match the XML files.
- `evaluate_build` got a pending status for "Paparazzi Sample App" but not "Paparazzi Sample Library" — the workflow still reports both.

Worth noting now that verification is real: on a failure Paparazzi writes the visual diff to `build/paparazzi/failures/debug/` (I confirmed both `<name>.png` and `delta-<name>.png` land there), and nothing in the workflow collects it. That's the artifact you'd actually want when a snapshot check goes red. Fine as a follow-up.

Remaining nits from last round, unchanged and entirely optional: the unused 518 KB `Newton.png` copy in the app module's debug assets, and the unreferenced `flix.svg` / `logo_flix.svg` at the sample root.

Nice work — approving. ✅

PR #305, opened 2026-08-13
Add Testify support for Paparazzi snapshots to Intellij plugin

DanielJette:

Add support for Test, Record, Pull, Delete, and GoTo Source/Image for [Paparazzi](https://github.com/cashapp/paparazzi) to the Intellij Platform Plugin.

https://github.com/user-attachments/assets/638b80da-8b12-48d2-ae1b-509ffb9d0dc2


Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/screenshot/ScreenshotPullAction.kt
```diff
@@ -24,21 +24,126 @@
...
 
-class ScreenshotPullAction(anchorElement: PsiElement) : BaseScreenshotAction(anchorElement) {
+class ScreenshotPullAction(anchorElement: PsiElement, testFlavor: TestFlavor) :
+    BaseScreenshotAction(anchorElement, testFlavor) {
 
-    override val classGradleCommand: String
-        get() = "screenshotPull"
+    override val gradleCommand: GradleCommand
+        get() = GradleCommand(
+            argumentFlag = "-PtestClass=",
```
AndroidTestifyBot: **Blocking — Testify regression.** No `$1` placeholder here, but `BaseScreenshotAction.toFullGradleCommand()` builds the argument with `argumentFlag.replace("$1", arguments)`. With nothing to replace, the test class is silently dropped and the command becomes:

```
:app:screenshotPull -PtestClass=
```

On `main` this was `-PtestClass=com.foo.MyTest#myMethod`. Should be `"-PtestClass=$1"` — or better, reuse `testFlavor.testGradleCommands.argumentFlag` so there's one source of truth.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/screenshot/ScreenshotClearAction.kt
```diff
@@ -25,18 +25,22 @@
...
 
-class ScreenshotClearAction(anchorElement: PsiElement) : BaseScreenshotAction(anchorElement) {
+class ScreenshotClearAction(anchorElement: PsiElement, testFlavor: TestFlavor) :
+    BaseScreenshotAction(anchorElement, testFlavor) {
 
-    override val classGradleCommand: String
-        get() = "screenshotClear"
+    override val gradleCommand: GradleCommand
+        get() = GradleCommand(
+            argumentFlag = "-PtestClass=",
```
AndroidTestifyBot: **Blocking — same `$1` omission as `ScreenshotPullAction`.** The command comes out as `:app:screenshotClear -PtestClass=` with an empty value, dropping the class filter that `main` passed.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/GoToBaselineAction.kt
```diff
@@ -27,21 +27,26 @@ import com.intellij.openapi.actionSystem.ActionUpdateThread
...
 import dev.testify.findScreenshotAnnotatedFunction
+import dev.testify.getElementAtCaret
 
 class GoToBaselineAction : BaseUtilityAction() {
 
     override fun getActionUpdateThread() = ActionUpdateThread.BGT
 
     override fun update(event: AnActionEvent) {
-        event.presentation.isEnabledAndVisible = event.findScreenshotAnnotatedFunction()?.let { function ->
+        val testFlavor = event.getElementAtCaret()?.determineTestFlavor() ?: return
```
AndroidTestifyBot: **Blocking — this always returns `null`, so Go To → Baseline Image (and ⌘⇧T) is dead.**

`getElementAtCaret()` is `psiFile.findElementAt(offset)`, which returns a *leaf* — `LeafPsiElement`, `PsiWhiteSpaceImpl` or `PsiComment`. `determineTestFlavor()` opens with `if (this !is KtElement) return null`, and no Kotlin PSI type is a leaf: I dumped all 310 classes in `org.jetbrains.kotlin.psi` from `kotlin-compiler-embeddable` and every one is composite (`KtElementImpl`, `KtElementImplStub`, `KtExpressionImpl`, …). None extends `LeafPsiElement`.

This regresses the Testify flavour too, which works on `main`. Suggest walking up first:

```kotlin
val testFlavor = event.getElementAtCaret()
    ?.parents?.filterIsInstance<KtElement>()?.firstOrNull()
    ?.determineTestFlavor() ?: run {
        event.presentation.isEnabledAndVisible = false
        return
    }
```

Separately: the bare `return` here (and on line 48) leaves `presentation` at whatever the previous `update()` set, so the item can stay enabled after the caret moves off a test.

Plugins/IntelliJ/src/main/kotlin/dev/testify/TestFlavor.kt
```diff
@@ -0,0 +1,133 @@
...
+        qualifyingAnnotations = setOf(PREVIEW_ANNOTATION),
+        isClassEligible = false, // TODO: This is just for now, eventually we may want class-level markers too
+        methodInvocationPath = { className, methodName -> "$className*$methodName" },
+        testGradleCommands = GradleCommand(
+            argumentFlag = "--rerun-tasks --tests '$1'",
+            classCommand = "validate${Variant}ScreenshotTest",
+            methodCommand = "validate${Variant}ScreenshotTest"
+        ),
+        recordGradleCommands = GradleCommand(
+            argumentFlag = "--updateFilter '$1'",
```
AndroidTestifyBot: **Blocking — `--updateFilter` no longer exists.** Per the [Compose Preview Screenshot Testing release notes](https://developer.android.com/studio/preview/compose-screenshot-testing-release-notes), the flag was removed in **0.0.1-alpha10** — the same release that introduced the `@PreviewTest` annotation `PREVIEW_ANNOTATION` matches on. So on any version where this flavour can be detected at all, Gradle fails with `Unknown command-line option '--updateFilter'`.

alpha10 also moved reference images to `{module}/src/screenshotTest{Variant}/reference/` with flat fully-qualified filenames (`com.sample.screenshottests.test1_da39a3ee_c2200e98_0.png`), which `findPreviewMethod()` doesn't match either — it expects the older nested `…/ClassName/method_hash_hash_0.png` layout. Worth pinning down the target version; there's no Compose-preview sample in the repo exercising this path.

Plugins/IntelliJ/src/main/kotlin/dev/testify/TestFlavor.kt
```diff
@@ -0,0 +1,133 @@
...
+            methodCommand = "update${Variant}ScreenshotTest"
+        ),
+        findSourceMethod = ::findPreviewMethod
+    )
+}
+
+fun PsiElement.determineTestFlavor(): TestFlavor? {
+    if (this !is KtElement) return null
+
+    if (this.hasPaparazziRule()) {
```
AndroidTestifyBot: **Blocking — performance.** This runs before any path check, so for every `KtClass` and every `KtNamedFunction` in *every* Kotlin file — production code included — on every daemon pass, the plugin opens a `KaSession`, walks the class hierarchy, resolves the type of every declared and inherited property, then walks each of those types' hierarchies looking for `TestRule`, up to 3 levels.

The line-marker providers used to bail on a cheap `virtualFilePath.contains("androidTest")` string test. Reordering restores that, and also fixes the false positive where a production class holding a `Paparazzi` field is classified as a Paparazzi test:

```kotlin
val path = containingKtFile.virtualFilePath
val flavor = TestFlavor.entries.find { "/${it.srcRoot}/" in path } ?: return null
if (flavor == TestFlavor.Paparazzi && !hasPaparazziRule()) return null
return flavor
```

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
```diff
@@ -88,7 +75,11 @@ abstract class BaseUtilityAction : AnAction() {
     protected fun findBaselineImage(currentFile: PsiFile, baselineImageName: String): VirtualFile? {
         val module = ModuleUtilCore.findModuleForPsiElement(currentFile)
         if (module != null) {
-            val files = FilenameIndex.getVirtualFilesByName(baselineImageName, module.moduleContentScope)
+
+            val files = findFilesByPartialNameOrRegex(
```
AndroidTestifyBot: **Blocking — substring matching can resolve the wrong baseline, and Delete then removes it.**

`findFilesByPartialNameOrRegex` matches with `file.path.contains(partialName)`. For a class `FooTest`, the path `…_SubFooTest_default.png` also contains `FooTest_default.png`. `files.first()` then picks an arbitrary element of an unordered index collection.

For Reveal / Go-To that's a wrong-file annoyance; for `DeleteBaselineAction` it's data loss, and the confirmation dialog is built from `baselineImageName` (the requested name), not the resolved file — so the user can't see the mismatch. Please anchor the match on a preceding `_` or `/`, and show the resolved path in the confirmation.

Also: `module` is now computed only as a null guard. `findFilesByPartialNameOrRegex` already takes a `scope` — passing `module.moduleContentScope` would restore the old scoping and cut the cost a lot, since this currently walks every image in the project, drawables included, on the EDT (`BaseFileAction.getActionUpdateThread() == EDT`).

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/screenshot/ScreenshotPullAction.kt
```diff
@@ -24,21 +24,126 @@
...
+    }
+
+    private fun handlePaparazziPull(event: AnActionEvent) {
+        val project = event.project ?: return
+
+        val module = ModuleUtilCore.findModuleForPsiElement(anchorElement)
+        val workingDirectory = module?.let { ExternalSystemApiUtil.getExternalProjectPath(it) }
+
+        if (workingDirectory == null) {
+            Notification("Android Testify", "Screenshot Pull", "Could not determine module path.", NotificationType.ERROR).notify(project)
```
AndroidTestifyBot: The deprecated `Notification(groupId, title, content, type)` constructor, with a group id that isn't declared in `plugin.xml`. The platform logs "Notification group … is not registered" and users get no settings entry for it. Add `<notificationGroup id="Android Testify" displayType="BALLOON"/>` and go through `NotificationGroupManager.getInstance().getNotificationGroup(...)`. Same on line 147.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/screenshot/ScreenshotPullAction.kt
```diff
@@ -24,21 +24,126 @@
...
+
+        val sourceFiles = sourceDir.walkTopDown()
+            .filter { it.isFile && it.extension == "png" }
+            .toList()
+
+        if (fileName.contains("*")) {
+            val regex = fileName.replace("*", ".*").toRegex()
+            sourceFiles.filter { regex.matches(it.name) }.forEach { file ->
+                val destFile = File(destDir, file.name)
+                Files.move(file.toPath(), destFile.toPath(), StandardCopyOption.REPLACE_EXISTING)
```
AndroidTestifyBot: `Files.move` bypasses the VFS, so the new baseline won't appear in the Project view or an already-open editor until something else triggers a refresh — and a stale `VirtualFile` may still point at the moved failure image. Follow up with `VfsUtil.markDirtyAndRefresh(true, true, false, destDir)`, or use the VFS API inside a write action.

`Files.move` also throws `IOException` straight out of `actionPerformed`, which surfaces as an IDE error dialog — worth catching and routing into the same notification. Both the walk and the move run on the EDT.

Plugins/IntelliJ/src/main/kotlin/dev/testify/PsiExtensions.kt
```diff
@@ -58,11 +75,23 @@ val AnActionEvent.moduleName: String
...
         val gradleModule = psiModule.replace(".", ":")
         return gradleModule
     }
 
+val AnActionEvent.selectedBuildVariant: String
+    get() {
+        val psiFile = this.getData(PlatformDataKeys.PSI_FILE) ?: return "Debug"
+        val module = ModuleUtilCore.findModuleForPsiElement(psiFile) ?: return "Debug"
+        val variant = AndroidFacet.getInstance(module)?.properties?.SELECTED_BUILD_VARIANT ?: "debug"
+        return variant.replaceFirstChar { if (it.isLowerCase()) it.titlecase(Locale.getDefault()) else it.toString() }
```
AndroidTestifyBot: `Locale.getDefault()` on what becomes a Gradle task name: under a Turkish locale a variant starting with `i` title-cases to `İ` and `verifyPaparazziİnternal` doesn't exist. `Locale.ROOT` for identifiers.

Line 91 is worth a second look too — `findModuleForPsiElement` on a test file returns the `…​.unitTest` source-set module, but `AndroidFacet` is attached to the holder module, so this likely returns null and every command falls back to `Debug`. `AndroidFacetProperties.SELECTED_BUILD_VARIANT` is also legacy; `GradleAndroidModel.get(module)?.selectedVariantName` is the current API.

Plugins/IntelliJ/src/main/kotlin/dev/testify/extensions/ScreenshotClassMarkerProvider.kt
```diff
@@ -42,24 +45,24 @@ class ScreenshotClassMarkerProvider : LineMarkerProvider {
...
-
-        val functions = PsiTreeUtil.findChildrenOfType(this, KtNamedFunction::class.java)
+    private fun KtClass.getLineMarkerInfo(testFlavor: TestFlavor): LineMarkerInfo<PsiElement>? {
+        if (testFlavor.isClassEligible.not()) return null
+        val functions: Set<KtNamedFunction> = PsiTreeUtil.findChildrenOfType(this, KtNamedFunction::class.java).filterNotNull().toSet()
         if (functions.isEmpty()) return null
-        if (functions.none(KtNamedFunction::hasScreenshotAnnotation)) return null
-
+        if (testFlavor.hasQualifyingAnnotation(functions).not()) return null
+        if ((testFlavor == TestFlavor.Paparazzi) && this.hasPaparazziRule().not()) return null
```
AndroidTestifyBot: `determineTestFlavor()` on line 48 already called `hasPaparazziRule()` to arrive at `TestFlavor.Paparazzi` — this repeats the full hierarchy + property-type resolution. Worth threading the earlier result through instead.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseFileAction.kt
```diff
@@ -30,9 +30,13 @@ import com.intellij.openapi.project.Project
 import com.intellij.openapi.util.IconLoader
 import com.intellij.openapi.vfs.VirtualFile
 import com.intellij.psi.PsiElement
+import dev.testify.TestFlavor
 import dev.testify.baselineImageName
 
-abstract class BaseFileAction(protected val anchorElement: PsiElement) : BaseUtilityAction() {
+abstract class BaseFileAction(
+    protected val anchorElement: PsiElement,
+    private val testFlavor: TestFlavor
```
AndroidTestifyBot: This is never read — a dead parameter threaded through `DeleteBaselineAction` and `RevealBaselineAction`. Either use it (picking the baseline naming scheme per flavour is arguably what the `findBaselineImage` matching problem needs) or drop it.

Plugins/IntelliJ/src/main/kotlin/dev/testify/FileUtilities.kt
```diff
@@ -0,0 +1,116 @@
...
+            findClassByName(className, project)?.let { psiClass ->
+                return findMethod(methodName, psiClass)
+            }
+        }
+    }
+    return null
+}
+
+/**
+ * Input: /Users/danjette/dev/android-testify/Samples/Paparazzi/src/test/snapshots/images/dev.testify.samples.paparazzi.ui.common.composables_CastMemberScreenshotTest_b.png
```
AndroidTestifyBot: Local absolute path in the KDoc. Also, this example uses the full package (`dev.testify.samples.paparazzi.ui.common.composables`) while the inline comment on line 72 says `composables` — the code relies on the former, so the comment below is misleading.

More substantively: `split("_")` with an exact part count means `fun renders_empty_state` or a nested test class (`Outer.Inner` → `Outer_Inner`) silently fails to navigate. Splitting from the right, or matching the class name and taking the remainder as the method, would be sturdier.

Samples/Paparazzi/src/test/java/dev/testify/samples/paparazzi/ui/common/composables/UnitTestExample.kt
```diff
@@ -0,0 +1,13 @@
+package dev.testify.samples.paparazzi.ui.common.composables
+
+import org.junit.Assert.assertEquals
+import org.junit.Assert.assertTrue
```
AndroidTestifyBot: Unused import — `ktlintCheck` will fail. The root build applies ktlint to `allprojects` with `ktlint_standard_no-unused-imports = enabled`, and `:PaparazziSample` is in `settings.gradle`.

This file and `BasePaparazziTest.kt` are also the only Kotlin files in the sample without the MIT licence header their siblings carry.

AndroidTestifyBot (changes requested):
## Review: Paparazzi & Compose Preview support

Nice generalization — pushing the per-flavor differences into a `TestFlavor` enum is the right shape, and `hasPaparazziRule()` handling the "rule wraps Paparazzi" pattern (including inherited fields) is a thoughtful touch that most tools get wrong. The `Variant` placeholder trick is clever, if a little cryptic.

That said, I found a handful of things that look functionally broken — including two regressions to the existing Testify behaviour — so I'm requesting changes rather than approving.

---

### Blocking

**1. `screenshotPull` / `screenshotClear` no longer pass `-PtestClass` (Testify regression)**

`ScreenshotPullAction.kt:48` and `ScreenshotClearAction.kt:36` declare `argumentFlag = "-PtestClass="` — no `$1` placeholder. `BaseScreenshotAction.toFullGradleCommand()` does `argumentFlag.replace("$1", arguments)`, so with no placeholder the class/method is silently dropped:

```
main:  :app:screenshotPull -PtestClass=com.foo.MyTest#myMethod
here:  :app:screenshotPull -PtestClass=
```

Both should be `"-PtestClass=$1"`, matching `TestFlavor.Testify`. (Or, better, have them reuse `testFlavor.testGradleCommands.argumentFlag` so there's one source of truth.)

**2. `GoToBaselineAction` is dead — `determineTestFlavor()` on the caret element always returns `null`**

`getElementAtCaret()` returns `psiFile.findElementAt(offset)`, which is a *leaf* PSI element — `LeafPsiElement`, `PsiWhiteSpaceImpl` or `PsiComment`. `determineTestFlavor()` opens with `if (this !is KtElement) return null`.

No Kotlin PSI type is a leaf. I dumped all 310 classes in `org.jetbrains.kotlin.psi` from `kotlin-compiler-embeddable` and checked their supertypes — every one is composite (`KtElementImpl`, `KtElementImplStub`, `KtExpressionImpl`, …); none extends `LeafPsiElement`. So the caret element is never a `KtElement`, and both `update()` (`GoToBaselineAction.kt:39`) and `actionPerformed()` (`:48`) bail immediately.

Net effect: **Go To → Baseline Image (and ⌘⇧T) stops working entirely**, including for the Testify flavour where it works on `main`. The rest of the code paths are unaffected because the line-marker providers pass a real `KtClass`/`KtNamedFunction`.

Something like this in `GoToBaselineAction`:

```kotlin
val testFlavor = event.getElementAtCaret()
    ?.parents?.filterIsInstance<KtElement>()?.firstOrNull()
    ?.determineTestFlavor()
```

Secondary bug in the same two spots: bare `return` in `update()` leaves `presentation` at whatever the previous invocation set. Always assign `isEnabledAndVisible = false` before returning.

**3. `--updateFilter` was removed from the Compose screenshot testing tool**

`TestFlavor.kt:105` emits `--updateFilter '<class>*<method>'` for `update{Variant}ScreenshotTest`. Per the [release notes](https://developer.android.com/studio/preview/compose-screenshot-testing-release-notes), that flag was removed in **0.0.1-alpha10** — the same release that introduced the `@PreviewTest` annotation this PR keys off of. So on any version where `PREVIEW_ANNOTATION` resolves, Gradle will fail with `Unknown command-line option '--updateFilter'`; the Record action can't succeed.

alpha10 also changed the reference-image layout to `{module}/src/screenshotTest{Variant}/reference/` with flat FQN filenames (`com.sample.screenshottests.test1_da39a3ee_c2200e98_0.png`). `findPreviewMethod()` assumes the older nested layout — parent directory = class name, `split("_").first()` = method name — so Go To Source likely won't resolve either. Worth pinning down which AGP/tool version the Preview flavour targets; there's no Compose-preview sample in the repo exercising this path (`Samples/Flix/src/screenshotTest` contains only `.DS_Store`).

**4. `findBaselineImage` substring match can resolve the wrong file — and Delete then deletes it**

`BaseUtilityAction.kt:79` now matches with `file.path.contains(partialName)`. Given a test class `FooTest`, the path `…_SubFooTest_default.png` also contains `FooTest_default.png`. `files.first()` then picks an arbitrary element of an unordered index collection.

For Reveal/Go-To that's a wrong-file annoyance; for `DeleteBaselineAction` it's data loss, and the confirmation dialog is built from `baselineImageName` (the *requested* name) rather than the resolved file, so the user can't see the mismatch. Please anchor the match — require a `_` or `/` immediately before it — and show the resolved path in the confirmation.

**5. Performance: `determineTestFlavor()` runs full symbol resolution on every element of every Kotlin file**

The line-marker providers used to bail on a cheap `virtualFilePath.contains("androidTest")` string test. Now `determineTestFlavor()` (`TestFlavor.kt:113-116`) calls `hasPaparazziRule()` **first**, before any path check — so for every `KtClass` and every `KtNamedFunction` in every Kotlin file (production code included), on every daemon pass, the plugin opens a `KaSession`, walks the class hierarchy, resolves the type of every declared and inherited property, then walks *each of those types'* hierarchies looking for `TestRule` — up to 3 levels of nesting.

Reordering fixes it, and also removes the false positive where a production class holding a `Paparazzi` field is classified as a Paparazzi test:

```kotlin
fun PsiElement.determineTestFlavor(): TestFlavor? {
    if (this !is KtElement) return null
    val path = containingKtFile.virtualFilePath
    val flavor = TestFlavor.entries.find { "/${it.srcRoot}/" in path } ?: return null
    if (flavor == TestFlavor.Paparazzi && !hasPaparazziRule()) return null
    return flavor
}
```

Related: `ScreenshotClassMarkerProvider.kt:57` calls `hasPaparazziRule()` a second time after `determineTestFlavor()` already computed it — thread the result through instead.

**6. Blocking `.get()` on a pooled read action from inside a read action**

`hasPaparazziRule()` (and its siblings) use `executeOnPooledThread { ReadAction.computeBlocking { … } }.get()`. Line-marker providers already run under a read action, so the calling thread holds the read lock while blocking on a second thread that needs to acquire it. With a write action pending, the pooled thread can't proceed and the caller can't release — a classic freeze. The pattern predates this PR (`hasScreenshotAnnotation` did the same), but #5 multiplies how often it's reached. Since the callers are already in a valid read/analysis context — `ScreenshotInstrumentationLineMarkerProvider` calls `analyze(this)` directly — the pooled hop can simply be dropped for these.

---

### Should fix

- **Unregistered notification group.** `ScreenshotPullAction.kt:105,147` use the deprecated `Notification(groupId, …)` constructor with a group id that isn't declared in `plugin.xml`. Add `<notificationGroup id="Android Testify" displayType="BALLOON"/>` and go through `NotificationGroupManager`, otherwise the platform logs "Notification group is not registered" and users get no notification settings entry.
- **Pull writes behind the VFS's back.** `Files.move` (`ScreenshotPullAction.kt:132,139`) bypasses the VFS, so the new baseline won't show up in the Project view or a already-open editor until something else triggers a refresh. Follow with `VfsUtil.markDirtyAndRefresh(...)`, or use the VFS API in a write action. `Files.move` also throws `IOException` straight out of `actionPerformed` → IDE error dialog; wrap it. Both the walk and the move happen on the EDT.
- **`selectedBuildVariant` probably always returns `Debug`.** `PsiExtensions.kt:91` resolves the module of the test file, which under module-per-source-set is `…​.unitTest`; `AndroidFacet` is attached to the holder module, so `getInstance()` likely returns null. Worth testing against a project with a non-debug variant selected. `AndroidFacetProperties.SELECTED_BUILD_VARIANT` is also legacy — `GradleAndroidModel.get(module)?.selectedVariantName` is the current API.
- **`Locale.getDefault()` on a Gradle task fragment** (`PsiExtensions.kt:92`). Under a Turkish locale a variant beginning with `i` title-cases to `İ` and the task name breaks. Use `Locale.ROOT`.
- **`<depends>com.intellij.modules.androidstudio</depends>`** makes the plugin uninstallable in IntelliJ IDEA + Android plugin. Is that intended? Nothing in the diff obviously needs it — `AndroidFacet` comes from `org.jetbrains.android`, already declared. If it is intended, it's a breaking compatibility change and belongs in `CHANGELOG.md` and the marketplace description.
- **ktlint will fail.** `UnitTestExample.kt:4` imports `assertTrue` unused; the root build applies ktlint to `allprojects` with `no-unused-imports` enabled and `:PaparazziSample` is in `settings.gradle`.
- **Missing licence headers** on `BasePaparazziTest.kt` and `UnitTestExample.kt` — the only Kotlin files in the sample without the MIT header their siblings carry.
- **`BaseFileAction.testFlavor` is never read** (`BaseFileAction.kt:38`) — a dead parameter threaded through `DeleteBaselineAction` and `RevealBaselineAction`. Either use it (picking the right baseline naming scheme per flavour is arguably what finding #4 needs) or drop it.
- **Dead imports in `BaseUtilityAction.kt`** after the refactor: `FileTypeManager`, `FileTypeIndex`, `FilenameIndex`, `GlobalSearchScope`. (`navigateToClass` is also unreferenced, though that predates this PR.)

---

### Nits

- `findPaparazziMethod` / `findTestifyMethod` (`FileUtilities.kt:55,73`) split on `_` and require an exact part count, so `fun renders_empty_state` or a nested test class (`Outer.Inner` → `Outer_Inner`) silently fails to navigate. Splitting from the right, or matching the class name and taking the remainder, would be sturdier.
- `FileUtilities.kt:70` tests `path.contains("/test")` (no trailing slash) while `determineTestFlavor` uses `"/${srcRoot}/"`. A checkout under a directory named `test-something` satisfies the former. `/src/test/`, or the module's actual source roots, would be safer.
- `FileUtilities.kt:67` embeds a local absolute path (`/Users/danjette/dev/…`) in the KDoc, and its example contradicts the inline comment a few lines below (`composables_…` vs the full FQN).
- `fileName.replace("*", ".*").toRegex()` leaves the package dots as regex wildcards. Cosmetic, but `Regex.escape` would make it exact.
- `ScreenshotPullAction`: `isPaparazziPullAvailable()` and `handlePaparazziPull()` duplicate the target resolution and directory walk — worth extracting. `update()` also re-walks `build/paparazzi/failures` on every invocation.
- `build/paparazzi/failures` and `src/test/snapshots/images` are hardcoded; both are configurable in Paparazzi. Constants plus a comment at minimum.
- `const val Variant = "\$Variant"` (`TestFlavor.kt:133`) reads as a template but is a literal a later `replace` fills in. `VARIANT_PLACEHOLDER` would match Kotlin const conventions and a one-line comment would save the next reader a double-take.
- `.filterNotNull()` on the `PsiTreeUtil.findChildrenOfType` result (`ScreenshotClassMarkerProvider.kt:54`) is a no-op.
- `FileTypeManager.getStdFileType("Image")` returns `UnknownFileType` rather than null if the images plugin isn't present, which would silently make `findFilesByPartialNameOrRegex` return nothing.
- `findBaselineImage` computes `module` and then only uses it as a null guard — the previous module scoping is gone. `findFilesByPartialNameOrRegex` already takes a `scope`; passing `module.moduleContentScope` would restore it and cut the search cost a lot (it currently walks every image in the project, drawables included).
- No tests. The module has JUnit and `TestFrameworkType.Platform` wired up but no `src/test`. The name-parsing helpers and `TestFlavor` selection are pure logic and cheap to cover — they'd have caught #1 and #4.
- `ComposableScreenshotTest` swapping `PaparazziTestRule()` for a raw `Paparazzi(...)` drops `appCompatEnabled = true` and the synchronous Coil loader, which is why the baseline changed. Reasonable if the point is to exercise the direct-field detection path — a comment saying so would stop someone "fixing" it later.
- `.gitignore`: the new blanket `.kotlin/` makes the existing `Plugins/IntelliJ/.kotlin/` entry redundant.
- `AGENTS.md` documents `./gradlew test`, but there are no tests.

Heads up: your working tree still has `Text(text = "Hello, Paparazzi!?")` plus a re-recorded PNG in `ComposableScreenshotTest` — presumably left over from exercising the Pull action. Worth reverting before the next push.

Plugins/IntelliJ/src/main/kotlin/dev/testify/PsiExtensions.kt
```diff
@@ -112,31 +163,167 @@ val KtClass.testifyClassInvocationPath: String
...
 fun AnActionEvent.getVirtualFile(): VirtualFile? =
     this.getData(PlatformDataKeys.VIRTUAL_FILE) ?: (this.getData(CommonDataKeys.NAVIGATABLE_ARRAY)
         ?.first() as? PsiFileNode)?.virtualFile
+
+fun KtElement.hasPaparazziRule(): Boolean {
+    val containingClass = (this as? KtClassOrObject) ?: this.parents.filterIsInstance<KtClassOrObject>().firstOrNull() ?: return false
+    return containingClass.hasPaparazziRule()
+}
+
+fun KtClassOrObject.hasPaparazziRule(): Boolean {
```
AndroidTestifyBot: **Should fix — this result is recomputed for every function, on every highlighting pass, with no caching.**

`determineTestFlavor()` is called once per `KtNamedFunction` by `ScreenshotInstrumentationLineMarkerProvider` and once per `KtClass` by `ScreenshotClassMarkerProvider`. For anything under a `/test/` source root it reaches this function, which climbs to the containing class and resolves it from scratch each time.

The cost per call isn't trivial: `hasPaparazziRule(classSymbol)` calls `fieldTypes(symbol)`, which walks the full class hierarchy and resolves the return type of every declared and inherited property, and then calls `isTestRule(fieldType)` on each result — which walks *another* full hierarchy per field type. On top of that, each call pays a thread hop through `executeOnPooledThread(...).get()`.

So a unit-test file with 20 test methods does 21 identical full resolves per pass, and every one of them produces the same answer, because they all resolve the *same* containing class. `da94a713` stopped this running over production code, which was the severe case, but `src/test/` holds all of a project's unit tests — not just the Paparazzi ones — so this is paid by every user with a `src/test/` source root regardless of whether they use Paparazzi at all.

There is no `CachedValue` anywhere in the plugin (`grep -rn CachedValuesManager Plugins/IntelliJ/src` → nothing), so this is the natural place to introduce one:

```kotlin
fun KtClassOrObject.hasPaparazziRule(): Boolean =
    CachedValuesManager.getCachedValue(this) {
        CachedValueProvider.Result.create(
            computeHasPaparazziRule(),
            PsiModificationTracker.MODIFICATION_COUNT
        )
    }
```

That collapses N calls per class to one per class per PSI modification, and it composes with whatever the deferred threading rework ends up doing rather than conflicting with it.

Plugins/IntelliJ/AGENTS.md
```diff
@@ -0,0 +1,59 @@
...
+
+*   **Build:** `./gradlew buildPlugin`
+*   **Run IDE:** `./gradlew runIde` (Starts a sandboxed Android Studio instance with the plugin installed)
+*   **Run Tests:** `./gradlew test` (`src/test/kotlin`; most of the plugin needs a live IDE, so
+    coverage is limited to logic that can be exercised without a `Project`)
+
+## Configuration
+
+*   **Plugin Version:** Defined in `gradle.properties` (`pluginVersion`).
+*   **Platform Version:** Defined in `gradle.properties` (`platformVersion`). Currently targeting Android Studio Otter (2025.3.1.2).
```
AndroidTestifyBot: **Should fix — this file is new in this PR but describes the codebase as it was *before* it.**

This line says the project targets **Android Studio Otter (2025.3.1.2)**. `gradle.properties` on this same branch says `platformVersion = 2026.1.4.4` — Quail 4 Canary 4. A contributor following this doc would target the wrong platform, which is precisely the failure mode the "Notes for Agents" section below warns about.

Two more staleness issues in the same file:

- Line 33 describes `ScreenshotInstrumentationLineMarkerProvider` as adding gutter icons "to test methods annotated with `@ScreenshotInstrumentation`". As of this PR it also matches `@org.junit.Test` (Paparazzi) and `@com.android.tools.screenshot.PreviewTest` (Preview), via `TestFlavor.qualifyingAnnotations`.
- There is no mention of the `TestFlavor` model at all, even though it is the central abstraction this PR introduces and the thing a future contributor most needs explained. `TestFlavor.kt` is listed only as a generic "Utility class" alongside `FileUtilities.kt`.

Since the whole point of the file is to orient someone who hasn't read the diff, it's worth landing it accurate or not landing it in this PR.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/screenshot/ScreenshotPullAction.kt
```diff
@@ -24,21 +24,170 @@
...
-    override val methodGradleCommand: String
-        get() = "screenshotPull"
-
     override val methodMenuText: String
         get() = "Pull screenshots for '$methodName()'"
 
     override val icon = "pull"
+
+    override fun actionPerformed(event: AnActionEvent) {
+        if (testFlavor == TestFlavor.Paparazzi) {
```
AndroidTestifyBot, line 68: Minor, and part of the deferred Preview work rather than a separate issue — but worth recording in the TODO on `TestFlavor.Preview` so it isn't rediscovered later.

This branches Paparazzi away from the base implementation, so `TestFlavor.Preview` falls through to `super.actionPerformed(event)` and runs the `gradleCommand` declared above — `:module:screenshotPull -PtestClass=…`. That is a Testify-only task; there is no `screenshotPull` on a Compose Preview module, so the invocation fails.

`update()` has the same shape: it only overrides `isEnabled` for Paparazzi, so for Preview the item inherits `isEnabledAndVisible = (project != null)` from `BaseScreenshotAction.update()` and is always offered.

`ScreenshotClearAction` is handled correctly — both nav handlers gate it behind `if (testFlavor == TestFlavor.Testify)`. Pull is the one that isn't gated.

AndroidTestifyBot (commented):
## Re-review at `c778c645`

**Disclosure first:** I wrote 14 of the 17 commits on this branch (`0bd4592a`..`c778c645`) in response to my own earlier review. That means this is not an independent review of most of the diff, and I'm not going to approve it — a reviewer signing off on their own patches is not a signal worth anything. What follows is what I verified, what actually changed, and two findings that are new.

---

### Blockers from [the previous review](https://github.com/ndtp/android-testify/pull/305#pullrequestreview-4927894719) — resolved

| # | Finding | Fix |
|---|---------|-----|
| 1 | `argumentFlag = "-PtestClass="` had no `$1`, so `toFullGradleCommand`'s `replace("$1", …)` was a no-op and pull/clear ran with an empty filter | `TESTIFY_TEST_CLASS_FLAG = "-PtestClass=$1"`, shared by all four call sites — `0bd4592a` |
| 2 | `getElementAtCaret()` returns a leaf, and `determineTestFlavor()` opened with `if (this !is KtElement) return null`. No Kotlin PSI type is a leaf, so Go To → Baseline Image and ⌘⇧T were dead for **all** flavours, Testify included | resolution starts from the nearest enclosing `KtElement`; `update()` also now clears the presentation before its early return — `9a1b771a` |
| 3 | `hasPaparazziRule()` ran before any path check, so every class and function in every Kotlin file — production code included — opened a `KaSession` and walked the hierarchy on each daemon pass | the source-root string check gates the resolve; the duplicate call in `ScreenshotClassMarkerProvider` is gone — `da94a713` |
| 4 | `findFilesByPartialNameOrRegex` matched on `path.contains`, so `FooTest_default.png` also matched `pkg_SubFooTest_default.png`, and `.first()` picked arbitrarily from an unordered index collection — which Shift-click delete then removed without confirming | match on file **name**, anchored on a `_` boundary, results sorted, `UnknownFileType` guarded, module scope with a project fallback — `09e9faf2` |

The rest of the review is addressed in `9b7f8ca2` (notification group registered — the deprecated `Notification(...)` constructor was logging *"group is not registered"*), `0e433446` (VFS refresh so pulled baselines appear without a manual refresh, `IOException` no longer escapes as an IDE error dialog, glob matching no longer leaves package dots as regex wildcards), `86d2c890` (build variant read from the holder module), `b5f70b68` + `b90ce1ac` (underscore-containing method names resolve; 8 unit tests over the parsers), `47a3f65a`, `a9dd616d`, `9817b799`, `b5c53c35`.

### One thing that wasn't in the review

The branch tip had been amended to fold in a baseline re-recorded from `Text(text = "Hello, Paparazzi!?")` while the source still read `"Hello, Paparazzi!"` — `:PaparazziSample:verifyPaparazziDebug` was failing on the branch as pushed. Restored in `c778c645`.

### Verified green at `c778c645`

- `./gradlew compileKotlin test buildPlugin verifyPluginStructure` → BUILD SUCCESSFUL; 8 tests, 0 failures, 0 skipped. Compile is warning-free (this PR had introduced two "No cast needed" warnings).
- `./gradlew ktlintCheck` at the repo root → BUILD SUCCESSFUL. This **failed** before `a9dd616d` on two unused imports; `Plugins/IntelliJ` is a separate Gradle build and is not covered by the root ktlint config, so only the sample was affected.
- `:PaparazziSample:verifyPaparazziDebug` → BUILD SUCCESSFUL.

---

### New findings

**1. `hasPaparazziRule` is recomputed for every element, every pass — should fix.** Details inline on `PsiExtensions.kt`. `da94a713` stopped this running over production code, but within a `/test/` source root it still resolves the full class hierarchy plus every field type once per function per highlighting pass, with a thread hop each time. There is no `CachedValue` anywhere in the plugin. This affects anyone with a `src/test/` source root whether or not they use Paparazzi.

**2. `AGENTS.md` documents the pre-PR codebase — should fix.** Details inline. It states the wrong platform version and describes the line marker providers as Testify-only, which is the thing this PR changes.

**3. Pull on a Preview test runs a Testify task.** Minor, inline on `ScreenshotPullAction.kt` — belongs with the deferred Preview work rather than being a separate issue.

### Open by decision, not by oversight

Both were assessed as high effort and deferred deliberately; both are recorded in code TODOs and under **Known limitations** in the CHANGELOG:

- **Compose Preview flavour.** `--updateFilter` was removed in Compose Preview Screenshot Testing 0.0.1-alpha10 — the same release that introduced the `@PreviewTest` annotation `PREVIEW_ANNOTATION` matches on — so record fails outright, and `findPreviewMethod` still assumes the pre-alpha10 nested reference layout. Verifying a fix needs a Compose Preview sample module, and there isn't one (`Samples/Flix/src/screenshotTest` holds only a `.DS_Store`). This is the one inline comment from the previous review that is still live.
- **The `executeOnPooledThread { ReadAction.computeBlocking { … } }.get()` helpers** in `PsiExtensions.kt`. Pre-existing, but worth restating now that `09e9faf2` moved `BaseFileAction` from `EDT` to `BGT`: `update()` on BGT runs under a read action, and blocking there on another thread's read-lock acquisition inverts the lock order if a write action is pending. `GoToBaselineAction` was already BGT on `main`, and this PR adds two more call sites on that path (`ScreenshotPullAction.update` → `paparazziScreenshotFileNamePattern`). Note that `ScreenshotInstrumentationLineMarkerProvider` calls `analyze {}` directly on the daemon thread and is fine, which is fairly good evidence the wrapper isn't needed at all.

### Not verified — needs a `runIde` smoke test

I could not confirm these by reading, and the manual pass has not been run. There is also no CI workflow covering `Plugins/IntelliJ` at all, so nothing else will catch these either:

- **`86d2c890`'s whole premise** — that `AndroidFacet` is absent on the `.unitTest` source-set module and present on the holder module. If that's wrong, the fallback is dead code and the variant was already resolving. Check with a non-debug variant selected.
- **`09e9faf2`'s module scope** — Paparazzi baselines under `src/test/snapshots/images` sit outside any source root, so `moduleContentScope` on the source-set module will likely miss them and the project-scope fallback is what actually does the work. Worth knowing which path fires.
- **`9a1b771a`** — ⌘⇧T and Go To → Baseline Image from inside a test method.
- **`0e433446`** — break a snapshot, run `verifyPaparazziDebug`, Pull, and confirm the file appears in the Project view with no manual refresh.
- Gutter icons on `ComposableScreenshotTest` (direct field) and `CreditStripScreenshotTest` (rule inherited from `BasePaparazziTest`), and **no** icon on `UnitTestExample`.

### Correction to my earlier review

Finding #4 claimed the delete confirmation dialog is built from the *requested* name. It isn't — `DeleteBaselineAction.kt:49` uses `virtualFile.presentableName`, the resolved file, so a mismatch would have been visible there. The wrong-file resolution bug was real and is fixed, but the genuinely silent paths were Shift-click, which skips confirmation entirely, and the menu label, which does use the requested name via `shortDisplayName(anchorElement)`.

---

### Verdict: **Comment**

Not *Request changes* — every blocker is resolved and the build is green. Not *Approve* — I wrote most of this diff, the PR is still a draft, and the four fixes whose premises rest on runtime platform behaviour have not been exercised in a sandboxed IDE. My read is that this is mergeable once the two should-fix items land and someone runs the smoke test above.

Plugins/IntelliJ/src/main/kotlin/dev/testify/TestFlavor.kt
```diff
@@ -0,0 +1,140 @@
...
+    ),
+
+    Paparazzi(
+        srcRoot = "test",
+        moduleFilter = ".unitTest",
+        qualifyingAnnotations = setOf(PAPARAZZI_ANNOTATION),
+        isClassEligible = true,
+        methodInvocationPath = { className, methodName -> "$className*$methodName" },
+        testGradleCommands = GradleCommand(
+            argumentFlag = "--rerun-tasks --tests '$1'",
```
AndroidTestifyBot: **Should fix — `--rerun-tasks` is build-wide, not task-wide.**

It tells Gradle to ignore up-to-date checks for *every task in the executed graph*, not just the test task. Measured against the sample in this repo:

```
$ ./gradlew :PaparazziSample:verifyPaparazziDebug --rerun-tasks --tests '…ComposableScreenshotTest*default'
BUILD SUCCESSFUL in 13s
65 actionable tasks: 65 executed
```

All 65 — KAPT stub generation, `compileDebugUnitTestKotlin`, asset merging, the lot — re-run to verify one snapshot. This sample is tiny; on an app-sized module this is the difference between a few seconds and a full rebuild, and it is paid on every single click of Test or Record.

The obvious narrowing does **not** work, which is worth knowing before you try it:

```
$ ./gradlew :PaparazziSample:verifyPaparazziDebug --rerun --tests '…'
BUILD SUCCESSFUL in 1s
65 actionable tasks: 65 up-to-date      # nothing re-ran, nothing was verified
```

`--rerun` only applies to tasks named on the command line, and `verifyPaparazzi$Variant` is a thin wrapper that delegates to `testDebugUnitTest` — so the flag lands on the wrapper and the actual test task stays up-to-date. Silently passing without verifying anything is worse than being slow.

Naming the underlying test task as well does work:

```
$ ./gradlew :PaparazziSample:verifyPaparazziDebug :PaparazziSample:testDebugUnitTest --rerun --tests '…'
See the Paparazzi report at: …
BUILD SUCCESSFUL in 4s
65 actionable tasks: 1 executed, 64 up-to-date
```

4.1s vs 13.5s, and it actually re-verifies. `BaseScreenshotAction` already resolves the variant to build the task name, so it has everything it needs to emit `test${Variant}UnitTest` alongside. Whatever shape you land on, please leave a comment saying why the second task name is there — it reads like a redundancy otherwise, and the next person will delete it.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
```diff
@@ -72,28 +54,29 @@ abstract class BaseUtilityAction : AnAction() {
...
-            val files = FilenameIndex.getVirtualFilesByName(baselineImageName, module.moduleContentScope)
-            if (files.isNotEmpty()) {
-                return files.first()
-            }
-        }
-        return null
+        val project = currentFile.project
+        val module = ModuleUtilCore.findModuleForPsiElement(currentFile) ?: return null
+
+        return findBaselineImageFiles(baselineImageName, module.moduleContentScope).firstOrNull()
```
AndroidTestifyBot: **Should fix — this replaced an indexed lookup with a full enumeration, on a path that runs in `update()`.**

Before this PR it was `FilenameIndex.getVirtualFilesByName(name, module.moduleContentScope)` — a hash lookup in the filename index. It is now `FileTypeIndex.getFiles(Image, scope)`, which returns *every image file in the scope*, then filters and sorts in Kotlin. And when the module-scoped pass comes up empty — which the KDoc above says is the normal case for Paparazzi, since its baselines sit outside any source root — line 79 does it again over `projectScope`.

So for a Paparazzi test this is two full image enumerations, and the second one covers every drawable, mipmap and asset in the project. `GoToBaselineAction.update()` and every `BaseFileAction.update()` call it, which is each time one of those menus is populated. `ActionUpdateThread.BGT` keeps it off the EDT but does not make it cheap, and a slow `update()` shows up as menu lag.

The suffix match is the right semantics — I am not asking you to give that up. But the exact-name case can stay indexed:

```kotlin
FilenameIndex.getVirtualFilesByName(baselineImageName, module.moduleContentScope).firstOrNull()
    ?: findBaselineImageFiles(baselineImageName, module.moduleContentScope).firstOrNull()
    ?: findBaselineImageFiles(baselineImageName, GlobalSearchScope.projectScope(project)).firstOrNull()
```

That makes every Testify lookup a hash hit and leaves the scan for the Paparazzi package-prefixed names only. If you want to go further, the Paparazzi name is derivable — `paparazziScreenshotFileName` already builds it — so that case could be an indexed lookup too, and the scan could disappear entirely.

Plugins/IntelliJ/src/main/kotlin/dev/testify/TestFlavor.kt
```diff
@@ -0,0 +1,140 @@
...
+        ),
+        findSourceMethod = ::findTestifyMethod
+    ),
+
+    Paparazzi(
+        srcRoot = "test",
+        moduleFilter = ".unitTest",
+        qualifyingAnnotations = setOf(PAPARAZZI_ANNOTATION),
+        isClassEligible = true,
+        methodInvocationPath = { className, methodName -> "$className*$methodName" },
```
AndroidTestifyBot: Why `*` rather than `.` here? I confirmed the generated filter works — `--tests 'dev.testify.samples.paparazzi.ComposableScreenshotTest*default'` selects exactly one test — so this is a question, not a defect report.

The cost of `*` is that it matches any run of characters, so the filter is looser than the click that produced it. Invoking Test on `a.b.FooTest#default` generates `a.b.FooTest*default`, which also matches `a.b.FooTest2.somedefault` and anything else whose FQ name starts with `a.b.FooTest` and ends with `default`. `a.b.FooTest.default` would be exact.

If there is a case that needs the wildcard — parameterised tests, or a Paparazzi name suffix I have not run into — a one-line comment here would settle it permanently. Otherwise `.` seems strictly better.

Plugins/IntelliJ/src/main/resources/META-INF/plugin.xml
```diff
@@ -8,10 +8,13 @@
     <depends>org.jetbrains.kotlin</depends>
     <depends>com.intellij.gradle</depends>
     <depends>org.jetbrains.android</depends>
+    <depends>com.intellij.modules.androidstudio</depends>
```
AndroidTestifyBot, line 11: This is the biggest user-visible change in the PR — every existing IntelliJ IDEA user loses the plugin at 5.1.0 — and I cannot find what forces it. It is called out in the README and CHANGELOG as a deliberate break, so I assume it is, but the reason is not recorded anywhere.

The only Android API the plugin touches is `org.jetbrains.android.facet.AndroidFacet` (`PsiExtensions.kt:105`), which comes from `org.jetbrains.android` — already declared on line 10, and available in IDEA. `platformType = AI` affects what you compile against, not what you can run in.

If something genuinely Studio-only is in play, worth a sentence in the CHANGELOG entry saying what. If it is only the build-variant resolution, `<depends optional="true" config-file="…">` would keep IDEA users on the Testify path they have today and degrade only variant detection — which already has a `Debug` fallback.

Plugins/IntelliJ/gradle.properties
```diff
@@ -1,12 +1,14 @@
...
 
 pluginGroup = dev.testify
 pluginName = Android Testify - Screenshot Instrumentation Tests
 pluginRepositoryUrl = https://github.com/ndtp/android-testify/tree/main/Plugins/IntelliJ
 # SemVer format -> https://semver.org
-pluginVersion = 5.0.0
+pluginVersion = 5.1.0
 
 # Supported build number ranges and IntelliJ Platform versions -> https://plugins.jetbrains.com/docs/intellij/build-number-ranges.html
 pluginSinceBuild = 242
```
AndroidTestifyBot, line 14: `pluginSinceBuild = 242` is unchanged by this PR, but 5.1.0 is already declaring a compatibility break, so it is the natural release to confirm or correct the floor.

The plugin compiles against 2026.1.4.4 (261) and leans hard on the Kotlin Analysis API — `KaSession`, `KaClassSymbol`, `KaVariableSymbol`, `expandedSymbol`, `declaredMemberScope`. That surface was renamed and reshaped across 242 → 243 → 251, and nothing here has ever been checked against a 242 IDE. If it does not resolve there, users on Ladybug get a `NoClassDefFoundError` at highlighting time rather than a clean "incompatible" message from the marketplace.

`pluginVerification { ides { recommended() } }` is already configured in `build.gradle.kts` — `./gradlew verifyPlugin` would answer this definitively. I did not run it here since it downloads IDEs.

Related, and the reason this has gone unnoticed: no CI workflow covers `Plugins/IntelliJ`. The repo has `central_release.yml` and `flix_sample.yml`, and neither touches it. A workflow running `test buildPlugin verifyPlugin` on changes under `Plugins/**` would have caught the AGENTS.md drift and would catch this.

Plugins/IntelliJ/src/main/kotlin/dev/testify/PsiExtensions.kt
```diff
@@ -112,31 +166,183 @@ val KtClass.testifyClassInvocationPath: String
...
+        CachedValueProvider.Result.create(
+            computeHasPaparazziRule(),
+            PsiModificationTracker.MODIFICATION_COUNT
+        )
+    }
+
+private fun KtClassOrObject.computeHasPaparazziRule(): Boolean {
+    val containingClass = this
+
+    return ApplicationManager.getApplication().executeOnPooledThread(Callable {
```
AndroidTestifyBot, line 228: Noting for the record rather than asking you to fix it here — we agreed this rework was deferred, and the caching added in `7ba39a15` is the right first step.

What changed is where this now sits. `computeHasPaparazziRule` is invoked from inside a `CachedValueProvider`, which the daemon calls while already holding a read action. So the shape is: daemon thread holds a read lock → blocks in `Future.get()` → pooled thread tries to acquire a read lock. Because the platform gives writers priority, a pending write action blocks the pooled thread, and the daemon cannot respond to cancellation because it is parked in a non-cancellable `get()` with no `checkCanceled` polling. That is the classic three-way stall.

The cache makes it far rarer — one compute per class per PSI modification instead of one per function per pass — so this is much less likely to bite than it was last round. But rarer is not never, and a freeze here presents as "Android Studio hangs when I open a Paparazzi test". Whenever the `executeOnPooledThread { ReadAction… }.get()` rework happens, this call site is the one that matters most, since it is the only one on the highlighting path.

Plugins/IntelliJ/src/main/kotlin/dev/testify/extensions/ScreenshotInstrumentationLineMarkerProvider.kt
```diff
@@ -28,6 +28,9 @@ import com.intellij.codeInsight.daemon.LineMarkerInfo
 import com.intellij.codeInsight.daemon.LineMarkerProvider
 import com.intellij.openapi.editor.markup.GutterIconRenderer
 import com.intellij.psi.PsiElement
+import dev.testify.TestFlavor
+import dev.testify.determineTestFlavor
+import dev.testify.getQualifyingAnnotation
 import org.jetbrains.kotlin.analysis.api.analyze
 import org.jetbrains.kotlin.name.FqName
```
AndroidTestifyBot: Unused now — the two `FqName(...)` comparisons this import served moved into `getQualifyingAnnotation`. Kotlin does not warn on unused imports, so a clean compile will not surface it (I checked: the forced rebuild produced no warnings at all).

Plugins/IntelliJ/src/main/kotlin/dev/testify/TestFlavor.kt
```diff
@@ -0,0 +1,140 @@
+/*
+ * The MIT License (MIT)
+ *
+ * Copyright (c) 2026 ndtp
+  *
```
AndroidTestifyBot: Stray leading space — should be ` *` to line up with the rest of the block. Same in the new `FileUtilities.kt:5`. It is inherited from an existing file that has the same typo (`BaseUtilityAction.kt`, `GoToBaselineAction.kt`), so this is a copy-paste artefact rather than anything you did, but these two files are new and could start clean.

Plugins/IntelliJ/src/main/kotlin/dev/testify/extensions/ScreenshotClassNavHandler.kt
```diff
@@ -68,12 +72,13 @@ class ScreenshotClassNavHandler(private val anchorElement: PsiElement) : GutterI
...
-            ScreenshotTestAction(anchorElement),
-            ScreenshotRecordAction(anchorElement),
-            ScreenshotPullAction(anchorElement),
-            ScreenshotClearAction(anchorElement)
-        )
+        val group = DefaultActionGroup()
+        group.add(ScreenshotTestAction(anchorElement, testFlavor))
+        group.add(ScreenshotRecordAction(anchorElement, testFlavor))
+        group.add(ScreenshotPullAction(anchorElement, testFlavor))
+        if (testFlavor == TestFlavor.Testify) {
```
AndroidTestifyBot, line 79: Not for this PR — a note for whenever a third flavour arrives.

`AGENTS.md` now says "adding support for another screenshot testing framework should mean adding a `TestFlavor` entry, not branching inside the actions", and I agree with that. But `testFlavor == TestFlavor.X` currently appears in four places: here, `ScreenshotInstrumentationAnnotationNavHandler.kt:82`, and `ScreenshotPullAction.kt:67` and `:76`.

The two nav handlers are the same predicate duplicated — a `supportsClear: Boolean` field on the enum would retire both and put the answer next to the other per-flavour facts. The `ScreenshotPullAction` pair is a genuine strategy difference (shell out to Gradle vs. move files locally) and is harder to express as data, so it is fine as a branch — but it would read better as an explicit `pullStrategy` field than as an `if` on the enum. Both are cheap now and get expensive once there are three entries.

Plugins/IntelliJ/src/test/kotlin/dev/testify/ImageNameParsingTest.kt
```diff
@@ -0,0 +1,103 @@
...
+    fun `a testify name is not mistaken for a paparazzi one`() {
+        assertNull(parsePaparazziImageName("OrientationTest_default"))
+    }
+
+    /**
+     * The counterpart to the `_`-anchored match in [findBaselineImageFiles]: the two names below
+     * belong to different classes, and a plain "contains" check would confuse them.
+     */
+    @Test
+    fun `a longer class name is a different test`() {
```
AndroidTestifyBot: This test does not test what its name and KDoc claim. The doc says it is "the counterpart to the `_`-anchored match in `findBaselineImageFiles`" — but the body only asserts that the two *parsers* return different class names, which is true regardless of how the matching works. If someone loosened `findBaselineImageFiles` to a plain `contains`, this test would still pass.

The matching rule is the part with the interesting edge, and it is reachable: `findBaselineImageFiles` is a top-level function, and its filter predicate is pure. Pulling that predicate out — `fun isBaselineNameMatch(fileName: String, target: String): Boolean` — would let this test assert the real claim, that `com.example_SubFooTest_default.png` does *not* match `FooTest_default.png` while `com.example_FooTest_default.png` does.

Same opportunity for `ScreenshotPullAction.matchesScreenshotName` — it is private, pure, and has the prefix/suffix-overlap edge that the `length >= prefix.length + suffix.length` guard exists for. Nothing in the suite covers it today.

The eight tests here are good, and the parsers were the right thing to pin down first. This is about the two functions that decide which file the destructive actions operate on.

AndroidTestifyBot (changes requested):
## Review — third pass, full re-read at `fc9a07cd`

The three threads left open last round are all closed: `hasPaparazziRule` is now cached per class against `PsiModificationTracker.MODIFICATION_COUNT`, `AGENTS.md` describes the branch rather than its ancestor, and removing the Preview flavour made the `ScreenshotPullAction` fall-through correct on its own. The `TestFlavor` model reads well and the KDoc added throughout is genuinely load-bearing — several of the non-obvious decisions (leaf resolution in `determineTestFlavor`, the `_`-boundary match, the holder-module fallback) are now explained where someone will actually find them.

### What I verified locally

| Check | Result |
|---|---|
| `compileKotlin test buildPlugin verifyPluginStructure` (forced, `--rerun-tasks --no-build-cache`) | BUILD SUCCESSFUL, 8 tests genuinely executed, 0 failures, no compiler warnings |
| `:PaparazziSample:verifyPaparazziDebug` | BUILD SUCCESSFUL — the re-recorded `ComposableScreenshotTest` baseline matches the new direct-`Paparazzi` rule |
| `--tests 'dev.testify.samples.paparazzi.ComposableScreenshotTest*default'` | selects exactly 1 test — `methodInvocationPath` produces a working filter |
| Paparazzi failure-output contract | Confirmed in `paparazzi-2.0.0-alpha05` bytecode: `ImageUtils.assertImageSimilar` writes **both** `delta-<name>.png` and the plain `<name>.png` into `failureDir`. The prefix match in `findPaparazziFailures` is correct, and the KDoc claim is accurate. |

So the feature works and the branch is green. What follows is one measured performance regression, one hot-path regression, and a set of questions and nits.

### Blocking

**1. `--rerun-tasks` rebuilds the entire task graph on every Test/Record invocation** (`TestFlavor.kt:90`, `:95`). Measured on the sample:

| Command | Tasks | Wall |
|---|---|---|
| `verifyPaparazziDebug --rerun-tasks --tests …` (what the plugin emits) | **65 executed** | 13.5s |
| `verifyPaparazziDebug testDebugUnitTest --rerun --tests …` | 1 executed, 64 up-to-date | 4.1s |

On a real app — where that graph includes KAPT, resource merging and full Kotlin compilation — the gap is much wider than 3×, and it is paid on *every* click of Test or Record. Details in the inline comment, including why plain `--rerun` on `verifyPaparazzi$Variant` silently does nothing.

**2. Baseline lookup regressed from an indexed name query to a full image enumeration** (`BaseUtilityAction.kt:78`). `FilenameIndex.getVirtualFilesByName` became `FileTypeIndex.getFiles(Image, scope)` + filter + sort, and on a miss it repeats the whole thing at **project scope** — in `update()`, on the BGT, every time one of these menus is populated. Android projects have thousands of images. Inline comment has a suggested fast path.

### Questions

- `plugin.xml:11` — what specifically requires `com.intellij.modules.androidstudio`? This drops IntelliJ IDEA support entirely, which is the largest user-visible change in the PR, and `AndroidFacet` already comes from `org.jetbrains.android`.
- `TestFlavor.kt:88` — why `*` rather than `.` as the class/method separator? It works, but it over-matches.
- `gradle.properties:14` — `pluginSinceBuild = 242` is unverified. `pluginVerification { ides { recommended() } }` is already configured and has never been run; no CI workflow covers `Plugins/IntelliJ` at all (the repo has only `central_release.yml` and `flix_sample.yml`). I did not run `verifyPlugin` here — it downloads IDEs.

### Nits

Unused import, a malformed licence-header line in the two new files, flavour branching spreading against `AGENTS.md`'s own guidance, and one test that doesn't test what its name claims. All inline, none blocking.

### Still unverified by anything

The demo video covers the happy paths, and I can see the pull and variant work in it. What it predates is the caching in `7ba39a15`: editing `BasePaparazziTest` must still flip the gutter icon on `CreditStripScreenshotTest`. That is the one behaviour the new `CachedValue` could plausibly get wrong, and it cannot be unit-tested here — the module has no platform test fixtures and `CachedValuesManager` needs a live `Project`. Worth one `runIde` pass before this leaves draft.

### Correction to my previous review

My last review reported the root `./gradlew ktlintCheck` as BUILD SUCCESSFUL. It is not — but that was the wrong thing to measure. All the failures are in `Samples/`, which this project does not hold to ktlint, and the set is byte-identical on `main`. The library source passes cleanly, and `Plugins/IntelliJ` is a separate Gradle build outside the root ktlint config. Net effect of this PR on lint: none.

Plugins/IntelliJ/AGENTS.md
```diff
@@ -0,0 +1,113 @@
...
+
+| Field | Purpose |
+| --- | --- |
+| `srcRoot` | Source root a test of this flavour lives under, e.g. `androidTest`. |
+| `moduleFilter` | Source-set module name suffix, e.g. `.unitTest`. |
+| `qualifyingAnnotations` | Fully qualified annotations that mark a test method. |
+| `isClassEligible` | Whether class-level gutter icons and menu actions apply. |
+| `methodInvocationPath` | Builds the class/method identifier passed to Gradle. |
+| `testGradleCommands` | The verify task and its argument flag. |
+| `recordGradleCommands` | The record task and its argument flag. |
```
AndroidTestifyBot, line 44: `cc2df9a9` gave `GradleCommand` a fourth field, `additionalTasks`, and this table still lists three.

It is the one field where the documentation genuinely matters: dropping `additionalTasks` does not fail loudly, it makes `--rerun` land on the lifecycle wrapper so Test and Record report success while verifying nothing. The KDoc on the field says so, but a reader who arrives via this table has no reason to open it.

The Paparazzi bullet below is now incomplete for the same reason — it says the flavour "Runs `verifyPaparazzi$Variant` / `recordPaparazzi$Variant`, scoped by `--tests`", where the emitted command also names `test${Variant}UnitTest`.

Plugins/IntelliJ/CHANGELOG.md
```diff
@@ -2,6 +2,25 @@
...
+      declared directly on the test.
+    - Pull moves failure images into the baseline directory and refreshes the IDE so they appear
+      immediately.
+    - The selected build variant is resolved from the holder module, rather than always producing a
+      `Debug` task name.
+- Fixed "Pull" and "Clear" dropping the `-PtestClass` filter for Testify tests.
+- Fixed "Go To Baseline Image" never resolving the test under the caret.
+- Fixed baseline lookup resolving to a similarly named test's image, which "Delete" could then
+  remove.
+- Test methods named with underscores now resolve for "Go To Source".
```
AndroidTestifyBot, line 22: Two user-visible changes from this round are missing here:

- **Test and Record no longer re-run the whole build graph.** Measured on the sample: 65 tasks / 13.5s → 1 task / 4s. On an app-sized module the old command included KAPT, resource merging and full Kotlin compilation on every click, so this is the most noticeable change in the release for anyone using the Paparazzi actions.
- **The `--tests` filter was over-matching.** `Class*method` is a Gradle glob, so invoking Test on `a.b.FooTest#default` also selected `a.b.FooTest2.somedefault`. That is a correctness fix, and it belongs next to the other "Fixed …" lines rather than being invisible.

The first is arguably the headline of 5.1.0 for Paparazzi users.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
```diff
@@ -72,28 +56,52 @@ abstract class BaseUtilityAction : AnAction() {
...
-                return files.first()
+        val project = currentFile.project
+        val module = ModuleUtilCore.findModuleForPsiElement(currentFile) ?: return null
+        val projectScope = GlobalSearchScope.projectScope(project)
+
+        val packageName = (currentFile as? KtFile)?.packageFqName?.asString().orEmpty()
+        val candidateNames = if (packageName.isEmpty()) {
+            listOf(baselineImageName)
+        } else {
+            listOf(baselineImageName, "${packageName}_$baselineImageName")
```
AndroidTestifyBot, line 93: The unprefixed Testify-shaped name is tried first for **both** flavours, and the collision it allows is not purely theoretical.

For a Paparazzi test, candidate 1 is `FooTest_default.png` and candidate 2 is `com.example_FooTest_default.png`. A project that holds a Testify test with the same simple class and method name — a repository mid-migration from Testify to Paparazzi, which is much of the audience for this PR — has a real `FooTest_default.png` on disk. Candidate 1 then hits and the Paparazzi test resolves to the Testify baseline. Reveal opens the wrong image; **Delete removes it**, and on the shift-click path without confirmation.

The KDoc above notes that Paparazzi baselines sit outside any source root, which makes this *more* likely rather than less: it means the module-scoped pass misses and both candidates are tried at project scope, where the Testify baseline is visible.

The fix is to order the candidates by flavour, putting the package-prefixed name first for Paparazzi. That needs the flavour here — which is the parameter I asked you to delete from `BaseFileAction` in round 3 as dead. It was dead at the time; this is the case that would have used it, so it is worth threading back in deliberately rather than reverting that commit.

Low severity and I am not blocking on it, but it is the same failure mode as the round-3 blocker: the wrong baseline resolved, then deleted.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
```diff
@@ -28,40 +28,24 @@ import com.intellij.openapi.actionSystem.AnAction
...
+import dev.testify.TestFlavor
 import dev.testify.baselineImageName
+import dev.testify.findBaselineImageFiles
 import dev.testify.getVirtualFile
-import com.intellij.openapi.module.ModuleUtilCore
 import org.jetbrains.kotlin.psi.KtFile
 
 abstract class BaseUtilityAction : AnAction() {
 
     override fun getActionUpdateThread() = ActionUpdateThread.EDT
```
AndroidTestifyBot: This is now inert. All three concrete paths override it to BGT — `BaseFileAction` explicitly, and `GoToBaselineAction` and `GoToSourceAction` individually — so the EDT declaration never takes effect.

Leaving it as the base default makes it a trap: the next subclass inherits EDT and does index lookups and `executeOnPooledThread(…).get()` on the UI thread, which is the deadlock shape already noted on `PsiExtensions.kt`. Since every subclass has independently concluded it needs BGT, the base should say BGT and the three overrides should go.

AndroidTestifyBot (approved):
## Review round 4 — approving

**Disclosure first: I wrote the five commits under review** (`cc2df9a9`…`4b4fe134`), so this is not an independent read of them. I have tried to compensate by verifying behaviour empirically rather than by re-reading my own reasoning, and by looking specifically for what those commits broke. Every new finding below is against my own work. Weigh the approval accordingly.

### The two blocking findings from [round 3](https://github.com/ndtp/android-testify/pull/305#pullrequestreview-4931840985) are fixed, and I checked them by breaking the sample rather than by inspection

I overwrote the committed baseline with a different image and ran the exact command the plugin now emits:

```
:PaparazziSample:verifyPaparazziDebug :PaparazziSample:testDebugUnitTest \
    --rerun --tests 'dev.testify.samples.paparazzi.ComposableScreenshotTest.default'
```

- **It failed**, as it must: `TEST-…ComposableScreenshotTest.xml` → `tests="1" skipped="0" failures="1" errors="0"`.
- `tests="1"` is also the round-3 separator question answered — the `.` form selects exactly one test.
- `65 actionable tasks: 1 executed, 64 up-to-date`.

The same corrupted baseline under the **old** command shape (`--rerun-tasks`, `Class*method`) fails identically at `65 actionable tasks: 65 executed`. So the change is 65 tasks → 1 with no loss of verification — which is the thing that had to be proven, since a `--rerun` that lands on the lifecycle wrapper alone reports success while verifying nothing.

Record was the path that could fail quietly, so I checked it the same way: with the baseline still corrupted, the emitted record command restored it to the exact original blob (`e817bbb8`), `1 executed`. It genuinely re-recorded, rather than falling through to verify mode.

One incidental confirmation: the failed run wrote **both** `dev.testify.samples.paparazzi_ComposableScreenshotTest_default.png` and `delta-…png`, into `build/paparazzi/failures/**debug**/`. That validates two things `findPaparazziFailures` assumes — `walkTopDown()` is needed because Paparazzi nests failures under a variant directory, and the start-anchored match is what keeps the `delta-` strip out of a class-level Pull.

Plugin build, forced (`--no-build-cache --rerun-tasks`, because a plain run reports `FROM-CACHE` and means nothing): `BaselineImageMatchingTest tests="10" failures="0"`, `ImageNameParsingTest tests="7" failures="0"`, no compiler warnings.

**ktlint does not apply to this PR at all** — `Plugins/IntelliJ` is a separate Gradle build with no ktlint plugin, and the samples are exempt by policy. This retracts the concern I raised in round 2; it was wrong then and is moot now.

### New, non-blocking — all introduced by my own commits

1. **`AGENTS.md` no longer describes `GradleCommand`.** `cc2df9a9` added a fourth field whose *removal makes Test and Record silently pass*, and did not update the table that exists to stop exactly that rediscovery.
2. **The changelog omits two user-visible fixes** from this round, including one correctness fix (`Test 'default()'` could previously also run `FooTest2.somedefault`).
3. **`findBaselineImage` tries the Testify-shaped name first for both flavours**, which can resolve a Paparazzi test to a Testify baseline in a project holding both. Narrow, but it is the same shape as the round-3 blocker.
4. **`BaseUtilityAction`'s `ActionUpdateThread.EDT` is now inert** — every concrete subclass overrides it to BGT — so it is a trap for the next one that doesn't.

### Pre-existing, not this PR, but worth an issue

When a Paparazzi snapshot mismatches, Gradle's report renderer throws `NoClassDefFoundError: org/gradle/reporting/CodePanelRenderer$Data` and the actual mismatch message never reaches the console. I reproduced it identically on the old command shape, so this PR neither causes nor worsens it — but it is the failure path of this PR's headline feature, and it is what a user will see in the Run window after clicking Test.

### Unchanged by agreement

The `com.intellij.modules.androidstudio` rationale (you confirmed the break is intended; the *reason* is still recorded nowhere), `pluginSinceBuild = 242`, the `executeOnPooledThread(…).get()` rework, and the flavour-branching refactor.

### Scope of what I verified

The IDE surface — gutter icons, the popup menus, Go To Baseline, Pull, and the `CachedValue`
invalidation from `7ba39a15` — is exercised through `runIde` by the author before review, so it is
covered; I am not treating it as an open gap. My contribution here is the Gradle-facing behaviour,
which I could break and re-check from the command line: that Test and Record still verify and still
record after the `--rerun` change, and that the filter selects one test.

The one thing I would still single out is `78ab9ddf`, since it changed how *every* baseline is
resolved and my confidence in it rests on reading `FilenameIndex` scoping rather than on running it.
Go To Baseline on both a Testify and a Paparazzi test is the check that settles it.

Approving — the blocking items are resolved and demonstrated, and the four residuals are
documentation and one edge case, none worth another round.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
DanielJette: Fixed

Plugins/IntelliJ/src/main/resources/META-INF/plugin.xml
DanielJette: This was added because the Jetbrains Marketplace runs auto-verifications against the plugin to authorize which flavor of Intellij is supported. Previously, the verifications would fail on the base Intellij product as it seems to lack the ability to filter based on requiring the Android plugin.

This plugin was always intended to be Android Studio-only. I refactored the plugin.xml and the build.gradle a few versions ago and that introduced this error of incorrectly declaring its support for base Intellij.

This change effectively reverts the previous requirement.

Plugins/IntelliJ/src/main/kotlin/dev/testify/actions/utility/BaseUtilityAction.kt
DanielJette: I'm going to fix this in a comprehensive follow-up.

There's already name collisions between modules within the Testify-only path anyway. For example, Flix and Legacy tests with the same filename will match and can take you/delete the wrong image already.

PR #306, opened 2026-08-14
5.1.1 - Remove `ReadAction.computeBlocking` usage

DanielJette:

Remove `ReadAction.computeBlocking` usage to prevent `Invocation of unresolved method` errors in Android Studio versions older than Quail | 2026.1

### DIFF
--- a/Plugins/IntelliJ/CHANGELOG.md
+++ b/Plugins/IntelliJ/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ## [Unreleased]
 
+## [5.1.1]
+
+- Remove `ReadAction.computeBlocking` usage to prevent `Invocation of unresolved method` errors in Android Studio versions older than Quail | 2026.1 
+
 ## [5.1.0]
 
 - **Breaking:** this plugin now requires Android Studio and can no longer be installed in IntelliJ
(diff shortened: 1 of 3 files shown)


AndroidTestifyBot (approved):
## Approving

Mechanical and well-scoped: seven call sites, all the same swap, plus the version bump and changelog entry. Nothing surprising.

Built it at `cb5e0c81` — `compileKotlin test verifyPluginStructure --no-build-cache --rerun-tasks` → BUILD SUCCESSFUL, 17 tests, 0 failures. `ReadAction.compute` resolves, so the substitution is real and not just plausible. The only remaining `computeBlocking` in the tree is the changelog prose describing the change.

This also settles something I flagged as unverified on #305: `pluginSinceBuild = 242` is load-bearing rather than stale. `computeBlocking` landed in 2026.1, so every build between 242 and 261 was hitting this. Good catch.

### One thing worth knowing, not worth blocking

`ReadAction.compute` is itself deprecated on the current target platform, so this takes the branch from **zero compiler warnings to seven**:

```
w: PsiExtensions.kt:125:24 'static fun <T : Any!, E : Throwable!> compute(...)' is deprecated. Deprecated in Java.
   ... and the same at 145, 160, 171, 229, 311, 333
```

That is the right trade — deprecated-but-present beats absent, and the whole point is to work on 242 — so I would not change it here. Flagging it only because "the plugin compiles clean" was true before this and is a useful signal to keep: it is worth either accepting the seven knowingly or suppressing them at the call sites, so the next real warning is not lost in the noise.

Worth noting the deadlock shape these sit inside is untouched by this PR and unchanged: `executeOnPooledThread { ReadAction.compute { … } }.get()` still blocks the caller on a pooled thread that needs the read lock. Neither variant of the API makes that better or worse — just recording that this PR neither introduces nor addresses it.

Trailing whitespace on CHANGELOG.md:9, which in Markdown is a hard line break. Almost certainly unintended, entirely cosmetic.

PR #308, opened 2026-08-16
Upgrades to major dependencies (java 25, kotlin 2.3.21, gradle 9.5.0, AGP 9.3.1)

DanielJette:

> [!CAUTION]
> **Critical Update Needed:** Check your configuration files. 
> This release uses Java 25 which requires you to update your app to use Java 25.
> Java 25 also requires Android Gradle Plugin 9.x — the published artifacts are Java 25 bytecode and cannot be consumed by AGP 8.x

- Upgrade kotlin to 2.3.21, gradle to 9.5.0 and AGP to 9.3.1
- Upgrade all project files to use Java 25
    - Java 25 requires you to update your app to use Java 25
    - Requires Android Gradle Plugin 9.x. The published artifacts are Java 25 bytecode (class file
      major version 69) and cannot be consumed by AGP 8.x
    - Gradle 9.5.0 is required to build; `kotlinOptions`, `lintOptions`, `packagingOptions` and the
      legacy `libraryVariants` API have been migrated to their AGP 9 replacements
    - **Removed** `ScreenshotScenarioRule.assertSame()` with an `ActivityScenario` context parameter
      (JVM name `assertSameContext`). Kotlin 2.3 makes it ambiguous with the no-arg `assertSame()`
      at any call site where an `ActivityScenario` is an implicit receiver, such as inside
      `ActivityScenario.test { }`, so it could no longer be called. Set the scenario explicitly
      instead: `rule.withScenario(scenario).assertSame()`


Library/build.gradle
```diff
@@ -124,6 +114,12 @@ android {
     }
 }
 
+// AGP 9 removed the legacy `libraryVariants` API that was previously used to set
+// `outputFileName`. The AAR is renamed by configuring its bundle task directly.
+tasks.withType(BundleAar).configureEach {
```
AndroidTestifyBot, line 121: This restores the AAR name, but `archivesBaseName` was also driving the **APK** name for the androidTest variant, and that part is now lost. Locally on this branch:

```
Library/build/outputs/apk/androidTest/debug/Library-debug-androidTest.apk
Ext/Ktx/build/outputs/apk/androidTest/debug/Ktx-debug-androidTest.apk
```

whereas `bitrise.yml` still hardcodes the old names:

```yaml
# bitrise.yml:219-220
- test_apk_path: "./Library/build/outputs/apk/androidTest/debug/testify-debug-androidTest.apk"
# bitrise.yml:254-255
- test_apk_path: "./Ext/Ktx/build/outputs/apk/androidTest/debug/testify-ktx-debug-androidTest.apk"
```

That's the current **Ktx** failure, and **Library** runs the same step.

Gradle 9 only removed the `archivesBaseName` *property*, not the concept — `base.archivesName` is the supported replacement and AGP still honours it. I verified via an init script that setting it restores the old name exactly (`testify-ktx-debug-androidTest.apk`):

```groovy
base {
    archivesName = pom.artifact
}
```

With that in place the AAR also goes back to `testify-<variant>.aar`, so the `BundleAar` block is still needed for the `-$version` suffix — but keeping `base.archivesName` means the APKs, and anything else deriving from the archive base name, don't silently change. Same applies to the other four library modules.

bitrise.yml
```diff
@@ -140,7 +140,7 @@ workflows:
     - git-clone@8: {}
     - set-java-version@1:
         inputs:
-        - set_java_version: '21'
+        - set_java_version: '25'
```
AndroidTestifyBot, line 143: Bitrise is updated, but the two GitHub Actions workflows still pin older JDKs and neither is in this diff:

- `.github/workflows/flix_sample.yml:28` → `java-version: 21`. This workflow triggers on `Samples/Flix/**`, so it ran on this PR and failed:
  ```
  > Task :FlixLibrary:compileDebugJavaWithJavac FAILED
  error: invalid source release: 25
  ```
- `.github/workflows/central_release.yml:14` → `java-version: 17`. This is the Maven Central publish job (`Library:publish`, `Plugin:publish`, …); it will fail the same way on the next release.

Both need `java-version: 25` (and `central_release.yml` is still on `actions/checkout@v2` / `actions/setup-java@v1` if you want to refresh it while you're in there).

Samples/Gmd/build.gradle
```diff
@@ -1,17 +1,16 @@
...
+import org.jetbrains.kotlin.gradle.dsl.JvmTarget
 
 plugins {
     id 'com.android.application'
-    id 'kotlin-android'
     id 'dev.testify'
 }
 
 android {
     compileSdk 33
```
AndroidTestifyBot: `compileSdk 33` no longer works under AGP 9. `:GmdSample:assembleDebug` fails on this branch:

```
> A failure occurred while executing CheckAarMetadataWorkAction
   > 1. Dependency 'androidx.databinding:viewbinding:9.3.1' requires libraries and applications
        that depend on it to compile against version 34 or later of the Android APIs.
        :GmdSample is currently compiled against android-33.
```

This is a regression from the AGP bump, not pre-existing: `viewbinding:8.13.2` declares `minCompileSdk=1`, `viewbinding:9.3.1` declares `minCompileSdk=34`. The module uses `viewBinding true`, so it's unavoidable — `compileSdk` needs to go to 34+ (and `targetSdk 33` is probably worth revisiting at the same time).

Note there's no `test_gmd` workflow in `bitrise.yml`, so CI stays green on this — it only shows up when someone builds the sample.

Separately: is jumping this sample from Java 1.8 straight to 25 intentional? It's the "old project" sample, so 1.8 may have been deliberate coverage for consumers who haven't migrated.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+- Upgrade kotlin to 2.3.21, gradle to 9.5.0 and AGP to 9.3.1
+- Upgrade all project files to use Java 25
```
AndroidTestifyBot, line 9: This undersells the consumer impact. I unzipped the AAR built from this branch — `classes.jar` entries are class file major version **69**, i.e. Java 25 bytecode. Consumers of `dev.testify:testify` will therefore need JDK 25 *and* an AGP whose D8/R8 accepts class file 69; anyone on AGP 8.x can't consume this release at all.

The 4.0.0 entry set the precedent for spelling this out:

> - Java 21 requires you to update your app to use Java 21

Suggest matching it, e.g.:

```markdown
- Java 25 requires you to update your app to use Java 25
- Requires Android Gradle Plugin 9.x — the published artifacts are Java 25 bytecode and cannot be consumed by AGP 8.x
```

And since `testify_version=6.0.0` is the current release, this presumably wants to land as 7.0.0 rather than a minor.

gradle/libs.versions.toml
```diff
@@ -12,12 +12,13 @@ dokka = "2.1.0"
...
+hilt = "2.60.1"
 junit = "1.3.0"
-kotlin = "2.2.10"
-kotlinPluginSerialization = "2.2.10"
+kotlin = "2.3.21"
+kotlinPluginSerialization = "2.3.21"
 kotlinxCoroutinesAndroid = "1.10.2"
-ktlint = "13.1.0"
+ksp = "2.3.11"
+ktlint = "14.2.0"
```
AndroidTestifyBot, line 21: Good catch bumping this — 13.1.0 was silently doing nothing under AGP 9, and I confirmed `:Library:runKtlintCheckOverMainSourceSet` actually executes again on 14.2.0.

One loose end: `Plugins/Gradle/build.gradle:11` still applies `id 'org.jlleitschuh.gradle.ktlint' version '13.1.0'` hardcoded, and bitrise runs `Plugin:ktlintCheck`. Worth pulling that from the catalog so the two can't drift. (Related: `Plugins/Gradle/gradle/wrapper/gradle-wrapper.properties` is still on 8.14 — harmless while it's an included build using the root wrapper, but stale for anyone building it standalone. Leaving the plugin itself at `JVM_17` looks right to me, since it has to run inside consumers' Gradle daemons.)

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,8 @@
+#Sun Aug 16 15:39:16 EDT 2026
```
AndroidTestifyBot: Machine-local timestamp from the `wrapper` task — worth dropping, it just creates churn on every wrapper update. (Adding `distributionSha256Sum` is a nice improvement, keep that.)

Library/src/main/java/dev/testify/scenario/ScreenshotScenarioRule.kt
```diff
@@ -505,6 +505,11 @@ open class ScreenshotScenarioRule @JvmOverloads constructor(
         } ?: throw ScenarioRequiredException()
     }
 
+    // Kotlin 2.3 added the CONTEXTUAL_OVERLOAD_SHADOWED diagnostic, which fires here because the
+    // no-arg `assertSame()` above takes precedence in overload resolution over this
+    // context-parameter variant. Suppressed to keep the Java 25 upgrade behaviour-neutral;
+    // resolving it for real would mean changing this public API.
+    @Suppress("CONTEXTUAL_OVERLOAD_SHADOWED")
```
AndroidTestifyBot: If the diagnostic is right that the no-arg `assertSame()` always wins overload resolution, then this context-parameter variant is unreachable from Kotlin call sites except via `assertSameContext` from Java. It has no callers anywhere in the repo — no tests, no samples, no docs reference it.

Suppressing is fine for keeping this PR behaviour-neutral, agreed. But since a breaking release is already on the table (see the CHANGELOG comment), this might be the moment to `@Deprecated` it instead and drop it in the next major, rather than carrying a suppressed diagnostic on dead API.

AndroidTestifyBot (changes requested):
Reviewed the AGP 9 / Gradle 9.5 / Kotlin 2.3 / Java 25 migration. The DSL migration itself is in good shape — `lintOptions`→`lint`, `packagingOptions`→`packaging`, `kotlinOptions`→`kotlin { compilerOptions }`, the `devices`→`localDevices` swap, dropping `kotlin-android` in favour of AGP's built-in Kotlin, and kapt→KSP all look correct, and the inline comments explaining the non-obvious AGP 9 replacements are appreciated. I verified locally on JDK 25 that `:Library:bundleReleaseAar` / `:Ktx:bundleReleaseAar` still produce `testify-6.0.0.aar` / `testify-ktx-6.0.0.aar`, and that `ktlintCheck` genuinely runs again on 14.2.0 (13.x was silently a no-op under AGP 9).

Three things need fixing before this merges:

**1. Removing `archivesBaseName` renamed the androidTest APKs and broke the instrumented-test CI steps.** The AAR name was preserved via the new `BundleAar` block, but the APK name wasn't. `:Ktx:assembleDebugAndroidTest` now emits `Ktx-debug-androidTest.apk` and `:Library:assembleDebugAndroidTest` emits `Library-debug-androidTest.apk`, while `bitrise.yml:219-220` and `bitrise.yml:254-255` still point at `testify-debug-androidTest.apk` / `testify-ktx-debug-androidTest.apk`. That's why the **Ktx** check is already red; **Library** will follow. Details and a verified fix in the inline comment.

**2. The GitHub Actions workflows were never bumped off Java 21/17.** `bitrise.yml` moved to 25, but `.github/workflows/flix_sample.yml` still pins `java-version: 21` and is failing on this PR with `error: invalid source release: 25`, and `.github/workflows/central_release.yml` still pins `java-version: 17`, which would break Maven Central publishing on the next release.

**3. `:GmdSample` no longer builds under AGP 9.** `androidx.databinding:viewbinding:9.3.1` declares `minCompileSdk=34` (8.13.2 declared `minCompileSdk=1`), and the sample is still on `compileSdk 33`, so `:GmdSample:assembleDebug` fails outright. No bitrise workflow covers GMD, so CI won't catch it.

Also worth deciding before release (non-blocking on this diff): the published AARs are now Java 25 bytecode — I confirmed `classes.jar` in `testify-6.0.0.aar` is class file major version 69. That forces every consumer onto JDK 25 *and* AGP 9, which is a larger breaking change than the CHANGELOG currently conveys. See the CHANGELOG comment.

The rest are minor notes inline.

AndroidTestifyBot (approved):
All seven points addressed. Re-verified on JDK 25 at `c355dade`:

| Was | Now |
| --- | --- |
| `Library-debug-androidTest.apk` / `Ktx-debug-androidTest.apk` | `testify-debug-androidTest.apk` / `testify-ktx-debug-androidTest.apk` — matches `bitrise.yml:219-220` and `:254-255`; AAR still `testify-6.0.0.aar` |
| `flix_sample.yml` on JDK 21, `central_release.yml` on JDK 17 | both on 25 (plus `checkout@v4` / `setup-java@v4`) |
| `:GmdSample:assembleDebug` failing on `checkDebugAarMetadata` | builds clean on `compileSdk 36` |
| `Plugin` pinning ktlint 13.1.0 inline | `alias(libs.plugins.ktlint)` at 14.2.0 |

`Plugin:test`, `Plugin:ktlintCheck`, `:Library:testDebugUnitTest` and `:Ktx:testDebugUnitTest` all pass locally with mockk 1.14.11, and all 15 CI checks are green — including the two emulator shards and the Library/Ktx instrumented-test steps that were consuming the misnamed APKs.

Two notes on the parts that went beyond the review:

**The `assertSame()` removal is better than what I suggested, and my premise was wrong.** I claimed the context-parameter overload had no callers; it did — `launchActivity<MainActivity>().test { rule.assertSame() }` in the Legacy sample resolved to it via the implicit `ActivityScenario` receiver, which is invisible at the call site. That makes this a genuine behavioural trap rather than dead API: had the diagnostic just been suppressed, that call would have kept compiling while silently binding to the no-arg overload. Removing it and rewriting the call as `rule.withScenario(this).assertSame()` is the right resolution, and the CHANGELOG entry explains the migration well. I checked the remaining `rule.assertSame()` sites in `Library/src/androidTest/**/scenario/` — they all use `.use { scenario -> … }` with a named parameter and call `withScenario(scenario)` first, so none of them were bound to the removed overload.

**Good catch on `Samples/Legacy`.** `defaultConfig { compileSdkVersion 35 }` was silently overriding the outer `compileSdk 34` through Groovy's owner delegation — removing it and setting `compileSdk 36` once is the correct fix, not just a version bump.

Non-blocking, for later:

- `:GmdSample` still has no bitrise workflow, so nothing guards the `compileSdk` fix. Given AGP 9 broke it once via a transitive `minCompileSdk`, a build-only workflow would be cheap insurance.
- `Plugins/Gradle/gradle/libs.versions.toml` keeps `gradle = "8.13.2"` and `kotlin = "2.2.0"`. I read that as deliberate — `compileOnly libs.gradle` plus a JVM 17 target keeps the plugin loadable in AGP 8 consumers' daemons even though the library artifacts now require AGP 9 — but flagging it in case the intent was to move it too.

LGTM.

PR #311, opened 2026-09-13
Fix RootViewWithoutFocusException being intermittently thrown by Espresso.closeSoftKeyboard()

DanielJette:
Refactor the closeSoftKeyboard() helper to work around a `RootViewWithoutFocusException`

This deliberately does *not* use `Espresso.closeSoftKeyboard`. Espresso resolves the view to act on through `RootViewPicker`, which selects a single root window *once* and then waits up to 10 seconds for *that* root to report window focus. It never re-selects.

When a test adds a second window — a dialog, a popup, an overflow menu — the platform hands focus over in two steps: the activity's window loses focus and, some milliseconds later, the new window gains it. In between, no window has focus. `RootMatchers.DEFAULT` only matches a dialog that *already* has window focus, but it matches the activity's own window regardless of focus, so a root pick performed during the hand-over latches onto the activity window. That window never regains focus, and the call fails with `RootViewWithoutFocusException` after stalling for 10 seconds. The hand-over window is tens of milliseconds wide even on an idle device, which is what makes the failure intermittent and load-dependent.

Espresso only ever needs the resolved view for its window token, which we can read from [activity] directly. Doing so keeps hiding the keyboard independent of window focus, and avoids paying for a root-view resolution on every screenshot — the keyboard is not showing in the vast majority of tests.

This method requires test-thread/ui-thread synchronization and so must be invoked from a background thread.

`InputMethodManager.hideSoftInputFromWindow` is deprecated in favour of `WindowInsetsController.hide(ime())`, but the insets controller offers no completion signal. The `ResultReceiver` overload tells us when the keyboard is actually gone, which a screenshot test needs to know before it captures. Espresso's own `CloseKeyboardAction` uses this same API for the same reason.


Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +98,70 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+        override fun onReceiveResult(resultCode: Int, resultData: Bundle?) {
+            hideAcknowledged.countDown()
+        }
+    }
+
+    /* True only when the keyboard was showing and a request to hide it was actually sent. */
+    val hideRequested = AtomicBoolean(false)
+    val requestSent = CountDownLatch(1)
+    activity.runOnUiThread {
+        val windowToken = (activity.currentFocus ?: activity.window.decorView).windowToken
```
AndroidTestifyBot: 🔴 **This always uses the activity window's token.** `hideSoftInputFromWindow` only acts when `windowToken` matches the window of the view the IME is currently serving (`servedView.getWindowToken() != windowToken` → reject). When a dialog that can use the IME has focus (plain `Dialog`, `DialogFragment`, `BottomSheetDialog`, Compose `Dialog`), the served view lives in the dialog's window, so the request is dropped and the keyboard stays on screen.

I reproduced this on an API 37 emulator: a `Dialog` with an `EditText` and the keyboard showing. This implementation took 2004 ms and the keyboard was still visible 1.5 s later. `Espresso.closeSoftKeyboard()` hid it in 313 ms, because `RootMatchers.DEFAULT` prefers the focused dialog.

Also, `currentFocus` is in the same window as `decorView`, so the `?:` doesn't change which token is used. Suggestion: on API 29+, choose the root that has window focus from `WindowInspector.getGlobalWindowViews()`, and fall back to the decor view.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +98,70 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+        if (windowToken != null) {
+            hideRequested.set(inputMethodManager.hideSoftInputFromWindow(windowToken, 0, resultReceiver))
+        }
+        requestSent.countDown()
+    }
+    requestSent.await(HIDE_KEYBOARD_TIMEOUT_SECONDS, TimeUnit.SECONDS)
+
+    /* Only wait for the keyboard to actually go away when there was one to dismiss. */
+    if (hideRequested.get()) {
```
AndroidTestifyBot: 🔴 **On API 36+ with `targetSdk >= 36`, this is always `true`.** Since Baklava, `hideSoftInputFromWindow` returns `true` even when the token check fails and nothing is sent (`ALWAYS_RETURN_TRUE_HIDE_SOFT_INPUT_FROM_WINDOW`), and in that case the `ResultReceiver` is never called. So the comment on L151 doesn't hold, and this `await` sits out its full 2 s timeout.

Measured on API 37 with `:Library`'s test APK (`targetSdk=36`): a plain `Dialog` on screen with **no keyboard** took **2002 ms** per call, versus 1–7 ms for the activity alone or an `AlertDialog`. The Legacy sample pins targetSdk 35, which is why CI doesn't show it.

Suggestion: don't rely on the return value. Check `WindowInsetsCompat.isVisible(ime())` on the target root first, and return right away if the keyboard isn't showing.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +98,70 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+ * [activity] directly. Doing so keeps hiding the keyboard independent of window focus, and avoids
+ * paying for a root-view resolution on every screenshot — the keyboard is not showing in the vast
+ * majority of tests.
+ *
+ * This method requires test-thread/ui-thread synchronization and so must be invoked from a
+ * background thread.
+ *
+ * [InputMethodManager.hideSoftInputFromWindow] is deprecated in favour of
+ * `WindowInsetsController.hide(ime())`, but the insets controller offers no completion signal. The
+ * `ResultReceiver` overload tells us when the keyboard is actually gone, which a screenshot test
```
AndroidTestifyBot: 🟡 The receiver doesn't tell us the keyboard is gone, and Espresso doesn't rely on it for that either.

- On API 36.1, when the token matches, `resultReceiver.send(...)` runs *before* `setImeVisibilityOnInsetsController(..., false)` starts the hide. On API 37, the IME insets animation was still running when this function returned, and ended about 36 ms later.
- Espresso's `CloseKeyboardIdlingResult.onReceiveResult` stays busy for **300 ms** after the callback, because "IMM responds to us first, before the messages are sent to the app that makes it draw over the region that was previously obscured by the keyboard". This implementation leaves that out.
- The 36.1 deprecation note on this overload says: "The ResultReceiver is not a reliable way of determining whether the Input Method is actually shown or hidden… use WindowInsetsController#hide… and verify the provided WindowInsets".

The idle sync after the call covered the animation in my runs, but that's down to timing. Waiting until the root's insets report the IME as hidden (or for `WindowInsetsAnimation.Callback.onEnd`) would give a real completion signal, and this paragraph should be updated to match.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +98,70 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+    val hideRequested = AtomicBoolean(false)
+    val requestSent = CountDownLatch(1)
+    activity.runOnUiThread {
+        val windowToken = (activity.currentFocus ?: activity.window.decorView).windowToken
+        if (windowToken != null) {
+            hideRequested.set(inputMethodManager.hideSoftInputFromWindow(windowToken, 0, resultReceiver))
+        }
+        requestSent.countDown()
+    }
+    requestSent.await(HIDE_KEYBOARD_TIMEOUT_SECONDS, TimeUnit.SECONDS)
```
AndroidTestifyBot: Nit: the timeout here, and `resultCode` in `onReceiveResult`, are both ignored without a trace. Espresso throws on timeout and on `RESULT_UNCHANGED_SHOWN`. Staying lenient is fine for screenshots, but a `Log.w` in either case would make "why is the keyboard in my baseline?" much faster to track down.

CHANGELOG.md
```diff
@@ -17,6 +17,9 @@
       at any call site where an `ActivityScenario` is an implicit receiver, such as inside
       `ActivityScenario.test { }`, so it could no longer be called. Set the scenario explicitly
       instead: `rule.withScenario(scenario).assertSame()`
+  - Fix RootViewWithoutFocusException being intermittently thrown by Espresso.closeSoftKeyboard()
+      Introduces a breaking change to EspressoHelper::closeSoftKeyboard() which now requires an
+      `Activity` parameter.
```
AndroidTestifyBot: A few things with this entry:
- At two-space indent it renders as a child of "Upgrade all project files to use Java 25". It's unrelated, so it should be a top-level `- ` item.
- The second line has no punctuation or break before it, so it runs on from the first as one sentence.
- `closeSoftKeyboard` is a top-level function (`dev.testify.internal.helpers.closeSoftKeyboard`), not a member of `EspressoHelper`, so `EspressoHelper::closeSoftKeyboard()` is misleading.

```suggestion
- Fix `RootViewWithoutFocusException` being intermittently thrown by `Espresso.closeSoftKeyboard()`
    - **Breaking:** `dev.testify.internal.helpers.closeSoftKeyboard()` now requires an `Activity` parameter
```

AndroidTestifyBot (changes requested):
## Summary

Getting away from `RootViewPicker` is the right call. I checked the root-cause analysis against the espresso-core 3.7.0 sources and it holds up. `RootMatchers.DEFAULT` is `(isDialog() && hasWindowFocus()) || isSubwindowOfCurrentActivity()`, and the second branch doesn't check focus. `RootViewPicker.waitForRootToBeReady` then waits up to 10s on the root it picked and never picks again. `CloseKeyboardAction` really does only use the view for `getWindowToken()`.

The replacement has two problems, though. I confirmed both on a device, and both show up in the exact situation this PR is about: a second window is on screen.

**Requesting changes** for the two blocking issues below.

---

## How I tested

I added a throwaway instrumentation test to `:Library` and called `closeSoftKeyboard(activity)` straight from the test thread. Device: API 37 emulator. `:Library`'s test APK uses `targetSdk=36`. I compared against `Espresso.closeSoftKeyboard()` in the same scenarios.

| Scenario | PR `closeSoftKeyboard(activity)` | `Espresso.closeSoftKeyboard()` |
|---|---|---|
| A. Activity only, no keyboard | 1–2 ms ✅ | – |
| B. Plain `Dialog` showing, no keyboard | **2002 ms** ❌ | – |
| C. `AlertDialog` showing, no keyboard | 3–7 ms ✅ | – |
| D. `EditText` in the activity, keyboard up | 3 ms, keyboard hidden ✅ (hide animation **still running** when the call returns) | 470 ms, hidden |
| E/F. Plain `Dialog` with an `EditText`, keyboard up | **2004 ms, keyboard still visible 1.5 s later** ❌ | 313 ms, hidden ✅ |

(If you want to rerun it, I can share the probe. It's about 150 lines, and `settings put secure show_ime_with_hard_keyboard 1` is needed so the emulator shows the soft keyboard.)

---

## 🔴 1. A keyboard in a focused dialog is no longer hidden

`hideSoftInputFromWindow` is a no-op unless the token you pass matches the window of the view the IME is currently serving. The framework has checked this on every API level: `servedView == null || servedView.getWindowToken() != windowToken` in both API 29 and 36.1. The PR always passes the **activity's** token. When a dialog that can use the IME has focus (a plain `Dialog`, `DialogFragment`, `BottomSheetDialog`, a Compose `Dialog`, or any dialog containing a text field), the IME is serving a view in the **dialog's** window. The call gets rejected and the keyboard stays up (row E).

Espresso handled this, because `RootMatchers.DEFAULT` picks the focused dialog first. So this is a regression for anyone taking screenshots of dialogs with input fields.

CI doesn't catch it because the Legacy sample's dialog tests use `MaterialAlertDialogBuilder`. AlertController sets `FLAG_ALT_FOCUSABLE_IM` on dialogs with no text input, so the IME stays with the activity (row C). The overflow-menu popup behaves the same way.

## 🔴 2. 2-second stall per screenshot on API 36+ when the app targets SDK 36

Since Baklava, for apps with `targetSdk >= 36`, `hideSoftInputFromWindow` **always returns `true`** (`ALWAYS_RETURN_TRUE_HIDE_SOFT_INPUT_FROM_WINDOW`). It returns `true` even when the token doesn't match and nothing is sent. In that case the `ResultReceiver` is never called. So `hideRequested` is `true`, `hideAcknowledged.await` runs until its 2s timeout, and the call returns silently. This happens in the same dialog scenarios as #1, **with or without a keyboard** (row B).

`targetSdk` in `libs.versions.toml` is 36, so consumers on current templates will hit this. The Legacy sample pins `targetSdkVersion 35`, which is the other reason CI looks fine.

## 🟡 3. The result callback doesn't mean the keyboard is gone

On API 36.1, when the token matches, the framework calls `resultReceiver.send(...)` *before* `setImeVisibilityOnInsetsController(viewRootImpl, false, ...)` starts the hide. Row D bears this out: the IME insets animation was still running when `closeSoftKeyboard` returned, and finished about 36 ms later. Espresso's own `CloseKeyboardIdlingResult` stays busy for another **300 ms** after the callback for this reason ("IMM responds to us first, before the messages are sent to the app that makes it draw…"). The PR leaves that part out.

In my runs, the `syncUiThread()` / `waitForIdleSync()` you added afterwards happened to cover the animation. That depends on timing, though, and the KDoc claim that the receiver "tells us when the keyboard is actually gone" and that Espresso uses it "for the same reason" isn't accurate. On 36.1 the overload is also deprecated specifically because "the `ResultReceiver` is not a reliable way of determining whether the Input Method is actually shown or hidden."

---

## Suggested direction

One approach that fixes all three and stays independent of window focus:

1. **Target the right window.** On API 29+, use the public `WindowInspector.getGlobalWindowViews()` and pick the root with `hasWindowFocus()`. Fall back to `activity.window.decorView` during the hand-over gap and on API 26–28.
2. **Skip early when there's nothing to hide.** If `ViewCompat.getRootWindowInsets(root)?.isVisible(WindowInsetsCompat.Type.ime())` is false, return without calling IMM. That keeps the common case free, and the Baklava return-value change stops mattering.
3. **Wait on the insets, not the callback.** Hide using that root's token, then poll (with a timeout) until the insets report the IME as hidden, or wait for `WindowInsetsAnimation.Callback.onEnd`. Stop relying on the return value and `ResultReceiver`.

Also, an instrumentation test in `:Library` covering rows B, D and E would guard against both regressions. Right now the only test changes are `every { closeSoftKeyboard(any()) } just runs` stubs, so none of the new logic is exercised.

## Smaller notes

- `activity.currentFocus ?: activity.window.decorView`: every view in the activity window has the same `windowToken`, so the `currentFocus` branch does nothing and suggests a different window could be chosen. (This goes away if you adopt suggestion 1.)
- `requestSent.await(...)`'s return value and the receiver's `resultCode` are both ignored. Espresso throws on `RESULT_UNCHANGED_SHOWN` and on timeout. Staying lenient is reasonable for screenshots, but a `Log.w` on timeout or on `RESULT_UNCHANGED_SHOWN` would make "why is the keyboard in my baseline" much easier to debug.
- CHANGELOG wording and nesting (inline comment).

The CI checks are green, and I checked that the new lines stay within ktlint's 120-column limit.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Confirmed and fixed. I reproduced it with the same setup on an API 37 emulator: plain `Dialog` + `EditText` with the keyboard up took **2002 ms and the keyboard was still visible 1.5 s later**, against 453 ms and hidden for `Espresso.closeSoftKeyboard()`. Your reading of the token check is right.

Rather than picking the focused root, the implementation now looks for the window whose insets report the IME as visible. That answers both "is there a keyboard?" and "which window owns it?" without consulting window focus at all, which felt worth preserving given what this PR is about — a dialog's keyboard is now hidden correctly even mid-hand-over. Same scenario after the change: **18 ms, hidden before the call returns**.

You're also right that `currentFocus ?: decorView` never changed the token. That's gone.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Confirmed and fixed. Reproduced at **2003 ms** for a plain `Dialog` with no keyboard showing, using `:Library`'s test APK (`targetSdk=36`) on API 37 — against 1–2 ms for the activity alone. Exactly as you describe: the return value is `true`, the `ResultReceiver` never fires, and the `await` sits out its full timeout.

The return value is no longer consulted on API 30+. The IME insets are checked first, so the call returns immediately when no keyboard is showing — that scenario is now **1 ms**. The `InputMethodManager` path survives only below API 30, where insets can't report IME visibility and the always-true behaviour doesn't apply.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Agreed — that paragraph overstated what the receiver gives you, and it's been rewritten.

On API 30+ the call now waits until the target window's insets stop reporting the IME as visible. In the dialog-with-keyboard scenario the keyboard is already gone when the call returns (insets report hidden immediately after, with no polling on the test side), so there is a real completion signal now rather than an acknowledgement that a request was received.

I deliberately left out Espresso's extra 300 ms: the insets flip at the end of the hide animation, and the caller follows with `syncUiThread()` / `waitForIdleSync()`, which loops the main thread until idle and covers the relayout of the uncovered region. The KDoc now states that explicitly instead of claiming the receiver means "keyboard gone".

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Done. Both paths now log a warning on timeout, naming the likely symptom ("It may be visible in the captured screenshot"), and the pre-API-30 path logs `RESULT_UNCHANGED_SHOWN`. Still lenient — a screenshot test shouldn't fail because the IME was slow — but no longer silent.

CHANGELOG.md
AndroidTestifyBot: Applied your suggestion verbatim. All three points were correct: it was rendering as a child of the Java 25 item, the second line ran on from the first, and `closeSoftKeyboard` is a top-level function rather than a member of `EspressoHelper`.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +105,178 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+        )
+    }
+}
+
+/**
+ * The root view of the window currently showing the soft keyboard, or null when no window is.
+ */
+@RequiresApi(Build.VERSION_CODES.R)
+private fun imeWindowView(activity: Activity): View? =
+    activity.rootWindowViews().lastOrNull { it.isImeVisible() }
```
AndroidTestifyBot: 🔴 **Reporting the IME in its insets doesn't make a window the IME's owner.** Two cases on API 37 where this picks the wrong root:

- **`AutoCompleteTextView` dropdown with the keyboard up.** `showDropDown()` sets `INPUT_METHOD_NEEDED`, so the popup is laid out behind the keyboard and reports `ime=true`. As the last root it wins. It has `FLAG_NOT_FOCUSABLE` and no IME control, so `hide(ime())` does nothing: **2010 ms, keyboard still visible**. `Espresso.closeSoftKeyboard()` hides it in about 400 ms.
- **Second activity over a first that had the keyboard up.** The stopped activity's root keeps stale insets (`ime=true`, `focus=false`) and is the only match: **2013 ms plus a spurious warning**, with no keyboard on screen.

Keeping only visible, focusable windows fixes both without relying on window focus. I ran exactly this change locally, and all 10 probe scenarios passed: activity, plain dialog, dialog + EditText, AlertDialog, both popup variants, AutoCompleteTextView, second activity. It needs `import android.view.WindowManager`.

```suggestion
private fun imeWindowView(activity: Activity): View? =
    activity.rootWindowViews().lastOrNull { it.canOwnIme() && it.isImeVisible() }

/**
 * Whether this root belongs to a window that can be the IME's target: one that is visible (not a
 * stopped activity, whose insets are stale) and focusable (not a popup, such as an
 * `AutoCompleteTextView` dropdown, that is laid out behind the keyboard and so reports it too).
 */
private fun View.canOwnIme(): Boolean {
    val flags = (layoutParams as? WindowManager.LayoutParams)?.flags ?: return false
    return windowVisibility == View.VISIBLE && (flags and WindowManager.LayoutParams.FLAG_NOT_FOCUSABLE) == 0
}
```

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +105,178 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+ * that it has gone.
+ *
+ * The insets are the platform's own answer to "is the keyboard showing?", which is why the
+ * deprecation note on [InputMethodManager.hideSoftInputFromWindow] points at them: the
+ * `ResultReceiver` overload acknowledges that the *request* was received, not that the keyboard has
+ * finished animating away.
+ *
+ * Note that the caller is still expected to sync the UI thread afterwards. The insets report the
+ * keyboard hidden at the end of its animation, but the app may still need to lay out and draw the
+ * region it used to cover.
```
AndroidTestifyBot: 🟡 The device doesn't match "the insets report the keyboard hidden at the end of its animation". I registered a `WindowInsetsAnimation.Callback` on the target root. In every hide case this function returned *before* the IME animation's `onEnd`:

| Case | Returned | `onEnd` |
|---|---|---|
| Activity `EditText` | 21 ms | 32 ms |
| `Dialog` + `EditText` | 62 ms | 65 ms |
| `AlertDialog` over keyboard | 37 ms | 51 ms |

The IME source is marked invisible when the hide *starts*. The idle sync that follows covered the remainder every time, so the conclusion (callers must sync afterwards) is right. It's the reason that needs changing. Suggested wording:

```suggestion
 * Note that the caller is still expected to sync the UI thread afterwards. The insets report the
 * keyboard hidden as soon as the hide begins, while its animation, and the app's relayout of the
 * region it used to cover, may still be in progress.
```

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +105,178 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+ */
+@RequiresApi(Build.VERSION_CODES.R)
+private fun closeSoftKeyboardViaInsets(activity: Activity) {
+    /* No window reports a visible IME, so there is no keyboard to dismiss. */
+    val imeWindowView = activity.onMainThread { imeWindowView(activity) } ?: return
+
+    activity.onMainThread {
+        imeWindowView.windowInsetsController?.hide(WindowInsets.Type.ime())
```
AndroidTestifyBot, line 183: 🟡 `androidx.core`'s `SoftwareKeyboardControllerCompat.Impl30.hide()` has a fallback for when the window can't control IME insets (their comment: multi-window mode, b/280532442). There, `WindowInsetsController.hide(ime())` does nothing, so they also call `imm.hideSoftInputFromWindow(view.windowToken, 0)`.

You now judge completion by polling the insets rather than by return values or the `ResultReceiver`, so making both calls here costs nothing and protects against a 2 s stall with the keyboard captured on tablets, foldables and desktop windowing. I couldn't reproduce multi-window on the emulator, so this is a robustness suggestion rather than a confirmed bug.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
```diff
@@ -87,6 +105,178 @@ class EspressoHelper(private val configuration: TestifyConfiguration) : Screensh
...
+}
+
+/**
+ * Hide the soft keyboard on API levels that predate [WindowInsets.Type.ime].
+ *
+ * Here the keyboard is hidden through the [InputMethodManager], which only acts when the token it
+ * is given belongs to the window hosting the view the IME is serving. We therefore target the
+ * focused window rather than the activity's. On these API levels the return value is meaningful —
+ * the change that makes `hideSoftInputFromWindow` always report success applies to API 36 and
+ * above.
```
AndroidTestifyBot: 🟡 "We therefore target the focused window rather than the activity's" is only true on API 29. On 26–28, `rootWindowViews()` returns just `window.decorView`, so a keyboard inside a dialog still isn't hidden there. That's a reasonable limitation, but worth saying here so nobody thinks those levels are covered. Also, as far as I can see this path hasn't been run on a device (the measurements in the commit are API 37 only). A quick run on the API 29 image would help.

AndroidTestifyBot (changes requested):
## Re-review of 0167e461

Finding the keyboard through the insets is a clear improvement. It sidesteps both the token check and the Baklava always-`true` return value, it leaves window focus out of it, and the no-keyboard case now costs one inset read. The CHANGELOG and logging feedback is addressed too, and CI is green. I've resolved those threads.

I re-ran the device probe against the new code with more scenarios, and found **one remaining blocking problem**. It's the same kind of bug as before: an extra window that *reports* the keyboard as visible gets picked even though it doesn't *own* it.

**Requesting changes** for that one issue. The fix is small, and I checked it on the device (below).

---

## How I tested

Same setup as before: API 37 emulator, `:Library` test APK at `targetSdk=36`, animations disabled, calling `closeSoftKeyboard(activity)` from the test thread. For each case I logged every root in `WindowInspector.getGlobalWindowViews()`: whether its insets report the IME, whether it has window focus, and whether it controls the IME insets.

| Scenario | 0167e461 | With the filter suggested below | `Espresso.closeSoftKeyboard()` |
|---|---|---|---|
| A. Activity only, no keyboard | 1 ms ✅ | 1 ms ✅ | |
| B. Plain `Dialog`, no keyboard | 0 ms ✅ | 0 ms ✅ | |
| D. `EditText` in activity, keyboard up | 21 ms, hidden ✅ | 21 ms, hidden ✅ | |
| E. `Dialog` + `EditText`, keyboard up | 62 ms, hidden ✅ | 36 ms, hidden ✅ | |
| J. `AlertDialog` over the activity's keyboard | 37 ms, hidden ✅ | 28 ms, hidden ✅ | |
| H1. Non-focusable `PopupWindow` over the keyboard | 22 ms, hidden ✅ | 15 ms, hidden ✅ | |
| **K. `AutoCompleteTextView` dropdown showing, keyboard up** | **2010 ms, keyboard still visible** ❌ | 31 ms, hidden ✅ | |
| **H2. `PopupWindow` with `INPUT_METHOD_NEEDED` over the keyboard** | **2013 ms, keyboard still visible** ❌ | 13 ms, hidden ✅ | 387–406 ms, hidden ✅ |
| **I. Second activity started while the first had the keyboard up** | **2013 ms + spurious warning** ❌ | 2 ms ✅ | |

---

## 🔴 `imeWindowView` picks windows that report the keyboard but don't own it

`lastOrNull { it.isImeVisible() }` treats "my insets include the IME" as "I own the IME". The two aren't the same. Here's what the root dump showed:

- **K / H2:** `AutoCompleteTextView.showDropDown()` sets `INPUT_METHOD_NEEDED` on its popup, and the platform lays that window out *behind* the keyboard. Its insets report `ime=true`, and as the most recently added window it wins `lastOrNull`. It can never be the IME target (it has `FLAG_NOT_FOCUSABLE`, and `focus=false imeCtrl=false`), so `hide(ime())` on it does nothing. The loop waits out its 2 s timeout and **the keyboard ends up in the screenshot**. Espresso handled this case. Screenshots of an autocomplete or exposed-dropdown menu with the keyboard up are a pretty ordinary thing to test.
- **I:** a stopped activity's root keeps its last insets. When the previous screen had the keyboard up, that root still reports `ime=true` (`focus=false imeCtrl=false`) after a new activity is on top. It's the only candidate, so it's chosen. Its insets never change, so the call takes 2 s and logs "may be visible in the captured screenshot", even though no keyboard is showing. This affects any test that moves from a form to another screen before calling `assertSame()`.

Suggested fix (inline): only consider roots whose window is **visible** and **focusable**. Neither check depends on window focus, so the hand-over gap is still handled. With exactly that filter applied locally, every row above passed (right-hand column).

---

## 🟡 Non-blocking

1. **Timing claim in the KDoc.** "The insets report the keyboard hidden at the end of its animation" didn't hold on the device. In every hide case the function returned *before* the IME's `WindowInsetsAnimation.Callback.onEnd`: D returned at 21 ms and the animation ended at 32 ms; E at 62/65 ms; K at 31/45 ms (with the filter). The insets flip when the hide *starts*. The `syncUiThread()` / `waitForIdleSync()` afterwards covered the rest in every run, so the behaviour is fine. Only the explanation needs correcting (inline).
2. **No fallback when the window can't control the IME insets.** `androidx.core`'s `SoftwareKeyboardControllerCompat.Impl30.hide()` calls `imm.hideSoftInputFromWindow(view.windowToken, 0)` when the window can't control IME insets (their comment cites multi-window, b/280532442), because `hide(ime())` alone does nothing there. Since you now poll the insets and ignore return values, adding that call is cheap insurance (inline). I couldn't test multi-window here.
3. **API 26–28.** `rootWindowViews()` only returns the activity's decor view below API 29, so the pre-API-30 KDoc's "target the focused window rather than the activity's" only holds on API 29. A keyboard in a dialog still won't be hidden on 26–28. That's an acceptable limitation, but the KDoc should say so. As far as I can tell, only API 37 has been measured, so the IMM path hasn't run on a device.
4. **Tests.** There's still no coverage of this logic: it's ~190 lines across two API paths, and so far every problem has come from device probing. Rows B, D, E, I and K would make a good `:Library` androidTest. I'm happy to share the probe as a starting point.
5. Nit: each poll goes through `onMainThread`, which can block for up to `HIDE_KEYBOARD_TIMEOUT_MILLIS` on its own. When the main thread is busy, the real ceiling is about twice the constant.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Verified on API 37: `Dialog` + `EditText` with the keyboard up now takes 62 ms and the keyboard is hidden. Resolving. The follow-up about which root gets picked (popups and stopped activities) is on L238.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Verified on API 37 / targetSdk 36: a plain `Dialog` with no keyboard now takes 0 ms. Resolving. The same 2 s symptom still shows up when an `AutoCompleteTextView` dropdown or a stopped activity's root is picked; that's tracked on L238.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Thanks for the rewrite. The new explanation (the insets flip at the end of the animation) didn't match what I measured, though: the function returns before the IME animation's `onEnd` in every hide case. Details and suggested wording on L167–169. I'm leaving this thread open until that's settled.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Verified. Resolving.

CHANGELOG.md
AndroidTestifyBot: Verified: the entry is top-level now and renders correctly. Resolving.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Confirmed and fixed. I reproduced the popup case exactly as you describe — an `INPUT_METHOD_NEEDED` popup over the keyboard left it on screen after the full 2 s wait — and the stale-insets case measured 2030 ms.

I've taken your filter. `canOwnIme()` keeps only visible, focusable windows, and as you say neither test touches window focus, so the hand-over gap is still handled.

I left the `AutoCompleteTextView` variant out of the test suite: its dropdown reliably SIGSEGVs the emulator's EGL driver on teardown (`libEGL_emulation.so`, `eglDestroySurface`, on the RenderThread), which is an emulator bug rather than anything to do with this code. The dropdown *is* a popup with `INPUT_METHOD_NEEDED`, so the explicit-popup test covers the same defect without the crash, and the KDoc names `AutoCompleteTextView` as the real-world instance.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: You're right, and I've taken your wording. I'd asserted the animation-end claim from the `syncUiThread()` that follows rather than from measurement, which was sloppy.

This one had teeth beyond the comment: because the insets flip early on API 30+, my test asserting immediately after the call passed there but failed on API 29, where I was measuring the visible display frame and the keyboard was genuinely still on its way out. The test now syncs and then allows the hide to finish, which is what a real caller does.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Added. I read through `SoftwareKeyboardControllerCompat.Impl30.hide()` and it does exactly what you describe, `b/280532442` and all.

I called `hideSoftInputFromWindow` alongside the insets controller rather than depending on `androidx.core` directly. `ViewCompat.getWindowInsetsController()` wouldn't have helped here anyway: on API 30+ it builds the compat from the controller alone, leaving `mView` null, which is precisely the branch that skips the IMM fallback. Getting it properly would mean adding `androidx.core` as a direct dependency of a published library, which seemed a lot to take on for a path neither of us can reproduce. Since completion is judged from the insets, the extra request is free where it isn't needed.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Good call — running it turned up a real bug. I booted an API 29 image and found that with a dialog open over a keyboard the path hid nothing: it aimed at the focused window, which is the dialog, while the keyboard still belonged to the activity behind it. The same mistake as the activity-token one, just one level down.

Since the return value is meaningful below API 36, every window is now offered the request until one is accepted, and the guess disappears. All 7 tests pass on API 29 and on API 37.

The API 26–28 limitation is documented as you asked, and it's now a genuine floor rather than an unknown: those levels can't enumerate windows at all, so the activity's own window is the only thing to offer.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Fixed — each poll now gets the time remaining rather than the whole constant, so the budget is what it says on the tin.

Library/src/main/java/dev/testify/internal/helpers/EspressoHelper.kt
AndroidTestifyBot: Agreed, and added: `Library/src/androidTest/.../CloseSoftKeyboardTest.kt`, covering the keyboard in an activity, in a dialog, under a dialog, under an `INPUT_METHOD_NEEDED` popup, and behind a stopped activity's stale insets, plus the two no-keyboard cases.

They're real regression tests, not just coverage — reverting only the source change gives 2 failures (`took 2030ms` and `still showing`). Every defect in this PR has come from device probing, as you say, so they now run on CI with the rest of `:Library`'s instrumentation. Verified on API 37 and API 29.

Thanks for pushing on this one; it found the API 29 bug too.

Library/src/androidTest/java/dev/testify/internal/helpers/CloseSoftKeyboardTest.kt
```diff
@@ -0,0 +1,331 @@
...
+        @BeforeClass
+        @JvmStatic
+        fun showKeyboardOnEmulators() {
+            shell("settings put secure show_ime_with_hard_keyboard 1")
+        }
+
+        @AfterClass
+        @JvmStatic
+        fun restoreKeyboardSetting() {
+            shell("settings delete secure show_ime_with_hard_keyboard")
```
AndroidTestifyBot, line 308: Optional: `settings delete` resets the setting to its default instead of putting back what was there before. Someone who has `show_ime_with_hard_keyboard` turned on for their own device or emulator will find it silently turned off after running `:Library`'s instrumentation tests. Reading the value in `@BeforeClass` (`settings get secure show_ime_with_hard_keyboard`, via the `ParcelFileDescriptor` that `executeShellCommand` returns) and writing it back here, or deleting only when it was `null`, would avoid that.

Library/src/androidTest/java/dev/testify/internal/helpers/CloseSoftKeyboardTest.kt
```diff
@@ -0,0 +1,331 @@
...
+     * sampling the instant it returns. A keyboard that was never asked to hide stays up, and still
+     * fails.
+     */
+    private fun assertKeyboardHidden(activity: Activity, dialog: Dialog? = null) {
+        getInstrumentation().waitForIdleSync()
+        assertTrue(
+            "The soft keyboard was still showing ${KEYBOARD_TIMEOUT_MILLIS}ms after closeSoftKeyboard()",
+            waitForKeyboard(activity, dialog, visible = false)
+        )
+    }
```
AndroidTestifyBot, line 246: Optional: the `*_hidesKeyboard` tests only check that the keyboard is *eventually* gone, up to 5 s after an idle sync. An implementation that waits out the 2 s timeout on the wrong window, while something else hides the keyboard, would still pass them. Only the two `returnsWithoutWaiting` tests catch stalls today. Timing the call in the hide tests too, e.g. `assertReturnedPromptly(timeCloseSoftKeyboard(activity))` in place of the bare `closeSoftKeyboard(activity)`, would cover both at no extra cost. I checked: every hide case finishes in under 70 ms on API 37.

AndroidTestifyBot (approved):
## Re-review of a13175ab — approving ✅

Everything blocking from the last two rounds is fixed, and I checked each fix on a device rather than taking it from the code. The new `CloseSoftKeyboardTest` also turns out to be a real regression test: it catches every bug this PR has had so far.

### What I verified

**1. The new test suite on two API levels**

| Device | Result |
|---|---|
| API 37 emulator (`targetSdk=36`) | 7/7 pass, 0 skipped |
| API 29 emulator (Android 10) | 7/7 pass, 0 skipped |

On API 29 this exercises the new offer-to-every-window `InputMethodManager` path, which none of the earlier device runs covered.

**2. The tests catch the earlier bugs.** I ran the new tests against each previous version of `EspressoHelper.kt`:

| Implementation under test | Failures |
|---|---|
| `f28aa40e` (activity token) | 2/7: `withKeyboardInDialog_hidesKeyboard` and `withDialogAndNoKeyboardShowing_returnsWithoutWaiting` |
| `0167e461` (unfiltered insets) | 2/7: `withInputMethodNeededPopupOverKeyboard_hidesKeyboard` and `withStaleInsetsFromStoppedActivity_returnsWithoutWaiting` |
| `a13175ab` (this PR) | 0/7 |

**3. My earlier 10-scenario probe, re-run on API 37.** It includes a real `AutoCompleteTextView` dropdown in addition to the synthetic popup.

| Scenario | a13175ab |
|---|---|
| Activity / plain `Dialog` with no keyboard | 0–2 ms |
| `EditText` in activity, keyboard up | 18 ms, hidden |
| `Dialog` + `EditText`, keyboard up | 24 ms, hidden |
| `AlertDialog` over the keyboard | 65 ms, hidden |
| Non-focusable popup / `INPUT_METHOD_NEEDED` popup over the keyboard | 23 / 26 ms, hidden |
| `AutoCompleteTextView` dropdown over the keyboard | 25 ms, hidden (Espresso: ~400 ms) |
| Second activity over one whose keyboard was up | 0 ms, no warning |

For what it's worth, I ran the real `AutoCompleteTextView` case three times on this emulator and never saw the EGL crash, and logcat from the latest run had no `Fatal signal`. Leaving it out of the suite is fine, though: the explicit popup test catches the same bug, as the mutation run above shows.

**4. The rest of the changes**
- `canOwnIme()` is the filter from last round. The pre-API-30 KDoc now states the API 26–28 limitation.
- Sending the request to every window with `any` is safe below API 36. A window whose token doesn't match is rejected on the client without side effects, and at most one window can match the view the IME is serving, so the shared `ResultReceiver` fires at most once.
- The extra `hideSoftInputFromWindow` call on API 30+ is harmless because its return value is ignored. That includes the always-`true` behaviour on 36+.
- KDoc timing wording is corrected. Each poll now gets only the remaining time.
- CI is green, including the `Library` instrumented job on the API 37 emulator. No line exceeds 120 columns.

### Optional follow-ups (inline, none blocking)
- `@AfterClass` deletes `show_ime_with_hard_keyboard` instead of restoring its previous value.
- The "hides keyboard" tests don't also check that the call returned quickly, so a future bug that stalls for 2 s but eventually hides would still pass.
- Minor: the two `onMainThread` calls before the poll loop still each use the full default timeout. So if the main thread is stuck, the worst case is about 4 s rather than 2 s. Not worth another round on its own.

Nice work pushing this all the way to device-verified tests.

PR #312, opened 2026-09-15
Explain how screenshots are compared

DanielJette:

Comparison is the most-searched topic in the docs (`diff`, `exactness`, `tolerance`, `threshold`, `fuzzy`, `delta e`), but the explanation is spread across four pages and a blog post, and searches such as `fuzzy` and `correctness` return nothing.

### How have you achieved it?

- New page `get-started/4-how-comparison-works.md`, placed between "Write a test" and "Update your baseline".
- States the exact per-pixel rule from `FuzzyCompare`: a pixel matches when `Delta E ≤ 100 × (1 − exactness)` (CIEDE2000), with a table of example values.
- Explains which comparison is chosen (`sameAsCompare`, `FuzzyCompare` or `compareMethod`) and notes that the Compose rules default to `exactness = 0.9`.
- Documents a caveat found in `HighContrastDiff`: when `exactness` isn't set, every non-excluded pixel in the diff image is black, so red and yellow only appear once `exactness` is set.
- Adds a tip on `6-verify-tests.md` linking to the new page.
- Adds `keywords` front matter with the terms people search for.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. All behaviour is taken from `TestifyConfiguration.getBitmapCompare()`, `FuzzyCompare`, `HighContrastDiff` and `AssertSame`.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested — fold into `6-verify-tests.md`

The content here is accurate and well-researched; the issue is placement, not correctness.

### 1. Merge into `6-verify-tests.md` and drop the new page

`6-verify-tests.md` already has `### Tolerance`, `### Exclusions` and `## Diagnosing Differences`, including a near-identical `defineExclusionRects` example and a byte-identical four-colour `Swatch` legend — the `Swatch` component is even redefined in the new file. Rather than a fourth get-started page plus a forward link, please fold the new depth (the exactness explanation, the Delta E formula and table, the diff-reading guide) into the existing page and drop `4-how-comparison-works.md`. That removes the duplication and leaves `4-` free in the sequence.

### 2. Qualify the exclusion-regions table row

The row reads:

> `exactness` set, **or exclusion regions defined**, or both → Excluded pixels are ignored, and remaining pixels are allowed a color difference set by `exactness`.

With exclusions and **no** `exactness`, `FuzzyCompare` takes the `else` branch — `baselinePixel == currentPixel` — so non-excluded pixels must match exactly, with no tolerance. Please add a clause along the lines of "…or required to match exactly if `exactness` isn't set".

### Verified correct

Everything else checked out, including some things that were easy to get wrong:

- `Delta E ≤ 100 × (1 − exactness)` matches `FuzzyCompare` exactly: `((100.0 - deltaE) / 100.0f >= exactness)`. The whole exactness table is arithmetically right.
- The comparison-selection logic matches `getBitmapCompare()`.
- `configure { }` beating the annotation matches `applyAnnotations`'s `if (exactness == null)`.
- Pass deletes the capture, failure keeps it and writes the diff — matches `AssertSame`.
- The caution that yellow and red only appear when `exactness` is set is **exactly** right: `HighContrastDiff`'s `else ->` branch paints `Color.BLACK` regardless of difference. Nicely caught.
- Compose rules defaulting to `0.9`, the size-mismatch early return, `.diff.png` naming, and all three `exclude*()` helpers all verified.

AndroidTestifyBot (approved):
## Approved

The restructure is exactly right. `4-how-comparison-works.md` is gone, the depth is folded into `6-verify-tests.md`, and the page reads as one document rather than two overlapping ones. The duplicated `Swatch` component is gone too — there's now exactly one definition, the existing one at the top of the file — and `4-` is free again in the get-started sequence.

The exclusion-regions row now reads:

> The remaining pixels are allowed a color difference set by `exactness`, or must match exactly if `exactness` isn't set.

which matches `FuzzyCompare` precisely: without `hasExactness`, `analyzePixelFunction` falls to `baselinePixel == currentPixel`.

### One improvement I didn't ask for

The size-mismatch sentence got sharper:

> Unless you provide your own `compareMethod`, two images of different sizes never match, whatever other settings you use.

That's more accurate than the original blanket claim. Verified both paths: `FuzzyCompare.compareBitmaps` early-returns `false` on a height or width mismatch, and `sameAsCompare` delegates to `Bitmap.sameAs()`, which returns false when dimensions differ. A custom `compareMethod` genuinely does bypass both. Good catch.

### Re-verified

- All four heading anchors resolve: `#diagnosing-differences`, `#writing-your-own-comparison`, and the inbound links from #319 and #320.
- Delta E formula and the full exactness table still match `FuzzyCompare`.
- The diff-colour caution still matches `HighContrastDiff`'s `else -> Color.BLACK` branch.

Docs build verified locally: clean build on `origin/main`, then with this branch applied — exit 0, zero warnings, and again in a combined build with the other six open PRs.

Issue #238, opened 2024-11-02
Improve error messages


PatrickVvB:
Hi!
In our project we using gradle composite builds. And this makes some problems with gradle tasks findings while sync the project. It is good when you have full understanding how composite build works and how other projects see path for another, but other cases some errors can frustrated.
For example, we have project like this: 
|-- foo-project
|     |-- composite-project
|     |     |-- feature-1
|     |     |-- feature-2
|     |           |-- domain
|     |           |-- ui
|     |-- settings.gradle.kts
|-- settings.gradle.kts
When we apply testify plugin on android library module it is try to find installDebugAndroidTest task using tasks.findByPath(moduleName + installDebugAndroidTest).
If we apply testify plugin on ui (feature-2:ui) module and define only applicationPackageId and testPackageId, fallback value for moduleName will be "ui" (project.name). tasks.findByPath cant find the tasks and return null and pass it to dependsOn, and we got 'A dependency must not be empty' exception.

In reality full task path will be `: composite-project:feature-2:ui:installDebugAndroidTest`. But if we wont to find this task inside composite-project (ui module build.gradle.kts), we must to pass `: feature-2:ui:installDebugAndroidTest` (as in define in composite-project-settings.gradle.kts)

This feedback relates to:
- [ ] The Kotlin library
- [ x ] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

Solution can be check finding task for null. And if it is null, throw exception with used parameters.

PR #313, opened 2026-09-15
Add a recipe for custom build types and product flavors

### LINKED ISSUE #131: How to use testify with custom build type?
Hey there,

in our app we have an custom build type for us, on which our instrumentation tests rely,, which is called GoogleMock, so build type and flavor. So the default installDebug isn't working.
I set this up like this:
```
extensions.configure<TestifyExtension>{
            this.installTask = "installGoogleMock"
            this.installAndroidTestTask = "installGoogleMockAndroidTest"
        }
```

This isn't working, since for some reason the androidx.test.runner.AndroidJUnitRunner cannot be found. The error looks like:
```
  No failed screenshots found
INSTRUMENTATION_STATUS: Error=Unable to find instrumentation info for: ComponentInfo{packageName.test/androidx.test.runner.AndroidJUnitRunner}
INSTRUMENTATION_STATUS: id=ActivityManagerService
INSTRUMENTATION_STATUS_CODE: -1
  Pulling screenshots:
  No failed screenshots found

```
However, if I use an example project, where I don't have the build type, same dependency versions and just use debug everything is working fine. So I guess either I have to configure something else or there might be an issue in testify, which leads to not working code in case of build variants. 
Look forward for help, thanks!

DanielJette:

Build types and product flavors have no docs page, and `flavor` has returned no results in docs search in 2023, 2024 and 2025. Two users in #131 worked out the configuration themselves and asked for it to be documented.

Fixes #131
Relates to #238

### How have you achieved it?

- New recipe `recipes/23-build-types-and-flavors.md`.
- Describes the inference in `GradleProjectExtensions.kt`: the first `install…Debug` / `install…DebugAndroidTest` task alphabetically, and `applicationId` plus the debug build type suffix (flavor suffixes are ignored).
- Groovy and Kotlin DSL examples setting `installTask`, `installAndroidTestTask`, `applicationPackageId` and `testPackageId` together.
- Notes that install tasks are task names, not paths, since the plugin resolves `:<moduleName>:<task>`.
- Verification with `testifySettings` and `-Pverbose=true`, and the nested-module workaround for #238.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. Behaviour checked against `TestifyExtension.kt`, `GradleProjectExtensions.kt` and `ScreenshotTestTask.kt`.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested

Accurate page — the inference table matches `GradleProjectExtensions` exactly, and the `testifySettings` sample output reproduces the real `println` block's field order and alignment. Two changes.

### 1. Lead with `moduleName` for nested modules

The nested-module section recommends running install tasks by hand. There's a better fix: the lookup is `findByPath(":${settings.moduleName}:${settings.installAndroidTestTask}")`, and `moduleName` is user-settable — so

```groovy
testify {
    moduleName "feature:login"
}
```

makes the path resolve to `:feature:login:installDebugAndroidTest` and restores the automatic dependency. It also feeds the `testifyModule` resValue and the `./gradlew $moduleName:screenshotPull` hints in error messages, which come out correct with the full path too.

Please make that the primary recommendation, with manual install tasks as the fallback.

### 2. Hedge the instrumentation error string

`INSTRUMENTATION_STATUS: Error=Unable to find instrumentation info…` is a platform message, not a Testify one, and it's what readers will paste into a search box. Please prefix it with "an error like this", the same treatment #321 gives its Hilt error.

### Verified correct

- `installTask` / `installAndroidTestTask` inference, including "first alphabetically" — `tasks.names` is a `SortedSet`.
- That the inferred package IDs exclude product-flavor suffixes — `applicationTargetPackageId` only reads `defaultConfig.applicationId` and the debug build type.
- `testifySettings` printing `applicationPackageId` as `targetPackageId`.
- `-Pverbose=true`, and that `screenshotTest`/`screenshotRecord` `dependsOn` the install tasks.
- Issue #238 is real, open, and exactly about this `findByPath` failure. The "skips installing without an error" claim is right — `getInstallDebugAndroidTestTask(project)?.let { }` swallows the null.

`flavorDimensions = ["backend"]` is the current AGP form and reads fine against an AGP 9 baseline — no change suggested unless you target an older minimum.

AndroidTestifyBot (approved):
## Approved

Both requested changes are in, and the `moduleName` section went further than I asked.

### `moduleName` now leads, with the mechanism explained

> The plugin looks for the install tasks at `:<moduleName>:<task name>`, and `moduleName` defaults to the module's own name, without its parent directories.

Verified: `TestifySettings.create` does `extension.moduleName ?: project.name`, and Gradle's `project.name` for `:feature:login` is `login`. So the inferred path really is `:login:installGoogleMockDebugAndroidTest`, which doesn't exist — and `getInstallDebugAndroidTestTask(project)?.let { }` swallows the null, hence the silent skip. Setting `moduleName "feature:login"` makes `findByPath` resolve. Manual installs are correctly demoted to a fallback.

The Groovy/Kotlin spelling is right in both tabs: `moduleName` has no `is` prefix, so both DSLs use the same name.

### The error string is now attributed, not just hedged

> The message comes from Android rather than Testify, and you'll see an error like this:

Better than what I asked for — telling readers the message isn't Testify's is more useful than a bare "an error like this".

### New claim, verified

> Testify runs the tests with `adb shell am instrument` using `testPackageId`, and reads screenshots from the device using `applicationPackageId`.

Correct on both halves: `ScreenshotTestTask` builds `.argument("$testPackageId/$testRunner")`, while `ScreenshotPullTask` and `DeviceUtilities` use `runAs(targetPackageId)` and interpolate `targetPackageId` into the SD card path. The follow-on — that a wrong `applicationPackageId` still lets tests run but sends `screenshotPull`/`screenshotClear` to the wrong app — follows directly and is right.

### Re-verified from the previous round

Inference table, "first alphabetically" (`tasks.names` is a `SortedSet`), flavor suffixes correctly excluded from the inferred IDs, `testifySettings` printing `targetPackageId`, `-Pverbose=true`, and issue #238.

PR #315, opened 2026-09-15
Add a recipe comparing screenshot storage options

DanielJette:

Test Storage is only explained inside the Gradle Managed Device recipe, and searches such as `useteststorageservice`, `teststoragedestination` and `shared` return nothing.

### How have you achieved it?

- New recipe `recipes/22-screenshot-storage.md` with a comparison table.
- On-device paths come from `DataDirectoryDestination`, `SdCardDestination` and the plugin's `DeviceUtilities.kt`.
- Documents that SD card storage takes priority when both options are set, and that `screenshotPull` / `screenshotClear` use `run-as`, so they need a debuggable build.
- Documents a gotcha: `TESTIFY_USE_SDCARD` only changes the arguments passed by `screenshotTest`. `screenshotPull` and `screenshotClear` read `testify { useSdCard }`, so the page recommends reading the variable inside the `testify` block.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. Behaviour checked against `Destination.kt`, `TestStorageDestination.kt`, `ScreenshotPullTask.kt` and `ScreenshotClearTask.kt`.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested

Strong page — the storage-path details are right in places that are easy to get wrong. One required change.

### 1. Bump `test-services` to 1.6.0, here and in `20-gmd.md`

The page specifies `androidTestUtil "androidx.test.services:test-services:1.5.0"`. That matches the existing `20-gmd.md`, so the PR is internally consistent — but `Samples/Gmd/build.gradle` uses **1.6.0**, so both doc pages trail the repo's own sample. Please update this page and `20-gmd.md` together, so the two don't end up disagreeing.

### Verified correct

- **App data path** `/data/data/<pkg>/app_images/screenshots/<device key>/` — `context.getDir("images", MODE_PRIVATE)` gives `app_images`, then `"$SCREENSHOT_DIR/<key>"`.
- **SD card path** `/sdcard/Android/data/<pkg>/files/testify_images/<device key>/` — and it correctly **omits** the `screenshots/` segment. `SdCardDestination.getOutputDirectoryPath` uses the bare `formatDeviceString(...)` unlike the data-dir version, and `DeviceUtilities.screenshotDirectory` agrees. That's a one-line difference in the source and the page gets it right.
- SD card wins when both flags are set — `getDestination()` checks `useSdCard()` first, and `addManifestPlaceholders` orders its `when` the same way.
- The `TESTIFY_USE_SDCARD` subtlety is **exactly** right: it's read via `"TESTIFY_USE_SDCARD".fromEnv(...)` in `ScreenshotTestTask` only, while `DeviceUtilities` reads `testifySettings.useSdCard` — so `screenshotPull`/`screenshotClear` genuinely do keep looking in the app data directory. Nicely caught.
- Test Storage writes to the data dir first and copies on `finalize()` — `TestStorageDestination : DataDirectoryDestination`. So "the copy in the app data directory remains" is correct.
- `SdCardDestinationNotFoundException` and `TestStorageNotFoundException` both exist, `run-as` is used by pull and clear, and the GMD output path matches `20-gmd.md`.

### Minor suggestion (not blocking)

The `useSdCard` Groovy and Kotlin DSL tabs are byte-identical, which invites readers to hunt for a difference that isn't there — and it weakens the signal in #316, where the same tab pattern carries a real difference (`recordMode` vs `isRecordMode`). Consider a single untabbed block.

AndroidTestifyBot (approved):
## Approved

Both items addressed:

- **`test-services` bumped to 1.6.0** in the new page *and* in `20-gmd.md`, so the two pages agree with each other and with `Samples/Gmd/build.gradle`.
- **The redundant DSL tabs are collapsed** into a single block with "This is the same in `build.gradle` and `build.gradle.kts`", which says the thing the tabs were failing to say. The now-unused `Tabs`/`TabItem` imports were removed too — no orphans left, the file imports only `OpenNew`.

### Re-verified

The storage details that were right the first time are unchanged:

- App data path `/data/data/<pkg>/app_images/screenshots/<device key>/` — `getDir("images", MODE_PRIVATE)` plus `"$SCREENSHOT_DIR/<key>"`.
- SD card path correctly **omits** the `screenshots/` segment, matching `SdCardDestination.getOutputDirectoryPath` and `DeviceUtilities.screenshotDirectory`. Still the detail I'd most expect a docs page to get wrong.
- SD card wins over Test Storage — `getDestination()` checks `useSdCard()` first.
- The `TESTIFY_USE_SDCARD` subtlety: read only in `ScreenshotTestTask`, so `screenshotPull`/`screenshotClear` genuinely keep looking in the app data directory.
- Test Storage writes to the data dir first and copies on `finalize()`.

PR #317, opened 2026-09-15
Add a recipe for blank or incomplete screenshots

### LINKED ISSUE #91: Screens which inflate views at a later time are never captured
**Describe the bug**
Regardless of which capture method I try, I can never actually capture anything on screen which either inflates views after creation or adds views to a recyclerview. Using `Screenshot.capture()` as a fourth capture method would fix this. `Screenshot.capture()` gives you access to a `bitmap` so comparing shouldn't been an issue.

This issue relates to:
- [x] The Kotlin library
- [ ] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**
Try to screenshot any activity with just a recyclerview into which you add views after `onCreate` of the ` Activity`.

**Expected behavior**
Screenshot is correctly taken.

**Screenshots**

**Desktop (please complete the following information):**
 - OS: MacOS

**Target Android Device (please complete the following information):**
 - Device: Pixel
-  Physical or Virtual: Android Virtual Device (Emulator)
 - API Level: 28
- Testify Key: [e.g. Output from `./gradlew testifyKey`]

**Additional context**

---

_Copied from @videogreg93 https://github.com/Shopify/android-testify/issues/245_

### LINKED ISSUE #93: Screenshot is taken before GlideImage drawable is loaded
**Describe the bug**
I'm not sure that this is a bug, but am wondering if there is recommended way for the screenshot to have to wait to be taken until after an image is loaded.

This issue relates to:
- The timing of when the library takes the screenshot

**To Reproduce**
I have previously been simply loading a drawable into a compose `Image` like this:
```
Image(
                painter = painterResource(id = imageModel.drawableResId),
                contentDescription = imageModel.contentDescription,
                contentScale = ContentScale.Fit,
                modifier = Modifier
                    .clip(CircleShape)
                    .size(dimensionResource(id = avatarSize.diameter))
            )
```
Running screenshot tests with this worked and gave the result on the left in the screenshots.
But we want to be able to switch over to use GlideImage to allow the image to be either loaded from a url OR a drawable like this (specifically, we are using this library https://github.com/skydoves/landscapist which uses https://github.com/bumptech/glide):
```
GlideImage(
            imageModel = avatarOverlayBadge.image.imageModel, // imageModel could be of type drawable or a String url
            ...
        )
```
This works in the app as expected and loads the image, however on the screenshot tests it seems now that the screenshot is taken before the image is loaded and gives the result on the right **most of the time**.  We did not change anything in the screenshot test and are still using a drawable, it just seems the library takes slightly longer to show the drawable than directly using a Compose `Image`.

I attempted to fix this using a `CountingIdlingResource` with something like:
```
val idlingResource = CountingIdlingResource("AvatarImageLoadCompletionResource", true).apply {
            increment()
}
IdlingRegistry.getInstance().register(idlingResource)
```

and then in the implementation code I decrement the idling resource on success of the image loading. This seemed to make the test work a rare amount of times.

**Expected behavior**
- I would expect the screenshot to look the same as the below image on the left

**Screenshots**
Screen Shot 2022-03-02 at 6 08 44 PM

- Testify Key:  29-1080x1920@441dp-en_US

**Additional context**
Any help would be very much appreciated thank you!!

---

_Copied from @jennatauro https://github.com/Shopify/android-testify/issues/264#issue-1157867026_

DanielJette:

The answers to #91 and #93 exist only in long issue threads. Docs searches for `async`, `wait`, `delay` and `animation` find nothing that explains when Testify captures.

Fixes #91
Fixes #93

### How have you achieved it?

- New recipe `recipes/25-blank-or-incomplete-screenshots.md`.
- Explains that Testify captures after `Espresso.onIdle()`, and that the Compose rules also wait for the composition to be idle.
- Idling resource pattern and the "never decrement" debugging tip from #93, including unique names per instance.
- Image loaders: resource loaded vs. frame drawn, tying the idling resource to the request listener, and disabling cross-fades and emulator animations.
- GPU-drawn content: `captureMethod = ::pixelCopyCapture` with a small `exactness`, from #91.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. Guidance comes from the #91 and #93 threads, updated to the current `configure { captureMethod = … }` API.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested

Good page in outline, and the capture-method half is accurate — but the timing half is scoped to the wrong rule, and the third-party examples need to be backed by something we actually compile.

### 1. The idle-wait claim only holds for `ScreenshotRule`

> Testify waits for your test's UI to be idle before it captures. It launches the activity, applies your view modifications and Espresso actions, then waits on `Espresso.onIdle()`.

That's an exact description of **`ScreenshotRule`**: `EspressoHelper.afterInitializeView()` runs `actions?.invoke()` then `syncUiThread()` → `Espresso.onIdle()`, registered via `addScreenshotObserver(espressoHelper)`.

**`ScreenshotScenarioRule` has none of it** — no `EspressoHelper`, no `onIdle()`, no `setEspressoActions`. This is deliberate: with `ActivityScenario` the test author already owns view automation, per [Android's own guidance on driving an activity to a new state](https://developer.android.com/guide/components/activities/testing#drive-activity-new-state), so Testify steps out of the Espresso mechanism entirely.

As written, the page promises an idle wait that doesn't happen on the newer rule — precisely where blank-screenshot problems are most likely.

**Please add a `ScreenshotScenarioRule` section** explaining that idling is the test author's responsibility there, and how to do it. `ScreenshotScenarioRuleExampleTests.setEspressoActions` in the Legacy sample is the pattern to follow — Espresso is driven directly inside `launchActivity { }.use { }` before `.withScenario(scenario).assertSame()`, so `perform()` does the synchronising:

```kotlin
launchActivity<TestHarnessActivity>().use { scenario ->
    scenario.onActivity { activity -> /* inflate / set up */ }

    Espresso.onView(withId(R.id.edit_text)).perform(typeText("Testify"))

    rule.withScenario(scenario).assertSame()
}
```

(Note the sample's method is *named* `setEspressoActions` but calls raw `Espresso.onView(...)` — there's no such API on the scenario rule.)

### 2. Please add the idling examples to the samples

The `CountingIdlingResource` block and the Glide 4 `RequestListener` don't exist anywhere in the repo, so nothing compiles them or keeps them honest — and Glide's listener signatures (nullability of `target`, presence of `isFirstResource`) shifted across 4.x minors while the page pins no minor version.

**Please add working versions of both to the sample suite** and have the page reference them, the way #318 references `FullscreenCaptureExampleTests.kt`. That's the difference between a recipe we can verify and one that rots.

### 3. The `decrement()` debugging tip needs scoping

> temporarily remove the call to `decrement()`. The test should now hang and then fail with an Espresso idling timeout

On `ScreenshotScenarioRule` nothing calls into Espresso, so the test won't hang — it captures immediately and the tip reports a false negative. Please scope it to `ScreenshotRule`.

### 4. Query on shadows and elevation

The GPU-drawn list ends with "and some shadows and elevation". The `useSoftwareRenderer` KDoc in `TestifyConfiguration` describes the *opposite* direction — that software rendering is what loses rounded corners, shadows and elevation. Is the mention right for the drawing-cache path, or bleeding in from the software-renderer caveat? As-is the two pages could contradict each other.

### Verified correct

- Default View capture really is the drawing cache — `createBitmapFromDrawingCache` does `view.isDrawingCacheEnabled = true; Bitmap.createBitmap(view.drawingCache)`.
- Both Compose rules default to `::pixelCopyCapture`, and both call `composeTestRule.waitForIdle()`, so "the Compose rules also wait for the composition to be idle" is right.
- Pairing PixelCopy with a small `exactness` is sound given `FuzzyCompare`.
- Fullscreen capture for other windows is consistent with `FullscreenCaptureMethod`.

PR #318, opened 2026-09-15
Add a recipe for capturing dialogs, menus and popups

DanielJette:

Docs searches for `dialog` and `popover` return no results. Fullscreen capture already handles anything in another window, but the docs never use the words people search with.

### How have you achieved it?

- New recipe `recipes/26-dialogs-menus-popups.md`.
- Views examples (dialog in `setViewModifications`, overflow menu with Espresso) are taken from `FullscreenCaptureExampleTests.kt`.
- Compose examples for `DropdownMenu` and `AlertDialog` using Material 3.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. The Compose `DropdownMenu` and `AlertDialog` examples were run with `screenshotRecord` in the Flix sample on an API 37 emulator, and the captured images show the menu and dialog. That test code was not committed.

### Notice

> [!WARNING]


AndroidTestifyBot (approved):
## Approved

The Views examples are near-verbatim reproductions of `Samples/Legacy/src/androidTest/java/dev/testify/sample/FullscreenCaptureExampleTests.kt` — same rule construction with `TestifyConfiguration(exactness = 0.95f)`, same `.captureFullscreen().configure { excludeSystemUi() }` chain, same `MaterialAlertDialogBuilder(it.context)` modification, same `openActionBarOverflowOrOptionsMenu(getInstrumentation().targetContext)` action. These are known-good, already-running examples rather than reconstructions, which is exactly the right provenance for a recipe page.

**Verified:**

- `captureFullscreen()` exists as `fun <TConfigurable : TestifyConfigurable> TConfigurable.captureFullscreen()` in `FullscreenCaptureMethod.kt`.
- `::fullscreenCapture` exists as a `CaptureMethod`-shaped function, so the alternative form in the Compose `AlertDialog` tab is valid too.
- `excludeSystemUi()` is a real `TestifyConfiguration` extension, alongside `excludeStatusBar()` and `excludeNavigationBar()`.
- `setViewModifications { }` takes `(rootView: ViewGroup) -> Unit`, so `it.context` resolves.
- `setEspressoActions { }` and `setComposeActions { }` both exist with the signatures used.
- "The Compose rules already use an `exactness` of `0.9` by default" — confirmed, `exactness: Float = 0.9f` on both `ComposableScreenshotRule` and `ComposableScreenshotScenarioRule`.
- Both linked sample files exist at the paths given (note the file is `FullscreenCaptureExampleTests.kt` plural while the class inside is singular — the link uses the correct filename).
- Ordering in the chains is safe: `captureFullscreen()` sets `configuration.captureMethod`, and a later `configure { excludeSystemUi() }` only adds an exclusion provider, so it doesn't clobber it.
- The advice to pair full-screen capture with `excludeSystemUi()` plus a tolerance is well-founded — the status and navigation bars are exactly the volatile regions, and `FuzzyCompare` is what handles the rest.

The "which elements need this" list at the top is a genuinely useful framing, and calling out that the Compose examples are Material 3 heads off the obvious copy-paste failure.

Docs build verified: `yarn build` passes clean, no broken links.

Issue #195, opened 2023-12-21
ScreenshotIsDifferentException: The path should not include an empty segment


DanielJette:
**Describe the bug**

This issue relates to:
- [ ] The Kotlin library
- [ ] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**

When a test fails, the following output is shown:

```
There was 1 failure:
1) default(com.foo.about.AboutScreenshotTest)
dev.testify.core.exception.ScreenshotIsDifferentException: 

*  The captured screenshot is different from the baseline screenshot.
*  Run `./gradlew app::screenshotPull` to view the differences.
*  Run `./gradlew app::screenshotTest -PtestClass=com.foo.about.AboutScreenshotTest#default` to run this test again.
```

Running the command `./gradlew app::screenshotPull` yields this error:

```
* What went wrong:
Cannot locate tasks that match 'app::screenshotTest'. The path should not include an empty segment (try 'app:screenshotTest' instead).
```

**Expected behavior**

The error message should only use a single `:` to delineate the module name

PR #319, opened 2026-09-15
Add an error reference page

DanielJette:

People paste exception names and stack traces into docs search (`ScreenshotBaselineNotDefinedException`, `MissingAssertSameException`, `com.shopify.testify…`), but no page lists Testify's exceptions.

Relates to #195, #238

### How have you achieved it?

- New page `docs/errors.md` (sidebar position 4), grouped by comparison, writing tests, locale, storage, capture, Compose, accessibility and the Gradle plugin.
- Each heading is the exception class name, so a pasted name lands on the right section.
- Messages and triggers were read from each exception class and its call sites.
- Notes the double colon in suggested commands (#195) and links `com.shopify.testify` stack traces to the migration guide.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested — one small fix, otherwise excellent

Coverage is **complete**. I enumerated every concrete exception class in `Library/`, `Ext/` and `Plugins/Gradle/` independently and found no omissions. (`TestifyException` is abstract and correctly left out.)

### 1. `sidebar_position` collides with #320

Both this page and `settings.md` (#320) declare `sidebar_position: 4`. Please change this page to **5**, and move `changelog.md` to 6 and `migration.md` to 7. Settings comes first.

While you're in there: `extensions/_category_.json` and `recipes/_category_.json` are *already* both at position 3 on `main` — worth fixing in the same pass.

### Verified correct

Several of these could easily have been wrong and weren't:

- **`ScreenshotTestIgnoredException` reports as skipped, not failed.** Right — `AssertSame` catches it and calls `Assume.assumeTrue(false)`, and `ErrorCause` maps `AssumptionViolatedException` → `"SKIPPED"`.
- **`FinalizeDestinationException` only occurs under `useTestStorage`.** Right, and non-obvious: `Destination.finalize()` is `= true`, and `TestStorageDestination` is the only override that can return false.
- **Issue #195 for the `app::screenshotPull` double colon.** Right — and worth noting the issue *title* is "The path should not include an empty segment", which sounds unrelated. The body is exactly this bug. Good citation.
- `IllegalStateException: Target view has 0 size` is a real string in both Compose rules.
- `dev.testify.compose.R.id.compose_container` is the exact ID `ComposableTestActivity` assigns its `ComposeView`.
- `RootViewNotFoundException`, `ScreenshotBaselineNotDefinedException` and `GradleExtensionException` descriptions all match their real messages, and the quoted device key matches the actual baseline folders in `Samples/Legacy`.
- The `src/debug` advice under `ActivityNotRegisteredException` matches the house convention used by `extensions/compose/1-setup.md` and `Samples/Flix`.

### Minor suggestion (not blocking)

`ImageBufferAllocationException` defers entirely to `LowMemoryException`. Consider one clause noting it reports the specific requested byte count, which helps readers tell the two apart in a stack trace.

AndroidTestifyBot (approved):
## Approved

The sidebar renumbering is complete and the ties are gone.

Verified the resulting order by building the site with all open PRs applied:

| Item | position |
|---|---|
| `intro.md` | 1 |
| `get-started/` | 2 |
| `extensions/` | 3 |
| `recipes/` | **3.5** |
| `settings.md` (#320) | 4 |
| **`errors.md` (this PR)** | **5** |
| `changelog.md` | 6 |
| `migration.md` | 7 |

No two items share a position any more — including the pre-existing `extensions`/`recipes` collision, which the `3.5` float resolves without renumbering anything downstream. Neat solution.

The `get-started/6-verify-tests.md#diagnosing-differences` link still resolves: #312 moves that content around but keeps the `## Diagnosing Differences` heading. Worth noting Docusaurus 2 doesn't validate anchors, so I checked that one by hand rather than relying on the build.

### Re-verified

The page body is unchanged, and coverage is still complete — every concrete exception class in `Library/`, `Ext/` and `Plugins/Gradle/` has a section. The three I most expected to be wrong are still right:

- `ScreenshotTestIgnoredException` reports as **skipped** — `Assume.assumeTrue(false)`, and `ErrorCause` maps it to `"SKIPPED"`.
- `FinalizeDestinationException` really is Test-Storage-only — `Destination.finalize()` is `= true` and only `TestStorageDestination` overrides it.
- The #195 citation is correct despite the issue's title describing a different problem.

Issue #200, opened 2024-01-06
Gradle project property useLocale is not used


DanielJette:
**Describe the bug**

useLocale has no effect

This issue relates to:
- [x] The Kotlin library
- [x] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**

`./gradlew FlixSample:screenshotTest -PuseLocale=en_CA`

**Expected behavior**

Should run tests using English/Canada locale. Runs tests in default device locale.

**Version**

2.0.0

Issue #199, opened 2024-01-06
Gradle extension property outputFileNameFormat is not used


DanielJette:
**Describe the bug**

Setting `outputFileNameFormat` does not impact the file name being used.

This issue relates to:
- [x] The Kotlin library
- [x] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**

```
testify {
    outputFileNameFormat = "n"
}
```

**Expected behavior**

Test files should be named `default.png`, but are still named `className_default.png`

**Version**

2.0.0

PR #314, opened 2026-09-15
Add a guide to running screenshot tests on CI

DanielJette:
### What does this change accomplish?

There's no CI page. Docs searches for `ci`, `github actions`, `azure`, `devops`, `bitrise`, `linux` and `firebase` often return nothing, and the Firebase Test Lab setup only exists in the #84 thread.

Fixes #84

### How have you achieved it?

- New recipe `recipes/24-continuous-integration.md`.
- Explains why the device key and rendering must match, and how to compare `testifyKey` with the `Baseline could not be found in <key>` message.
- GitHub Actions and Bitrise examples are trimmed from this repo's own `flix_sample.yml` and `bitrise.yml`.
- Firebase Test Lab section based on #84, using `--environment-variables useSdCard=true` (read by `getDestination()` as an instrumentation argument) and `--directories-to-pull`.
- Sharding uses AndroidJUnitRunner's `numShards` / `shardIndex`, as `flix_sample.yml` does. It deliberately doesn't use `-PshardCount` / `-PshardIndex`: `ScreenshotTestTask` reads them with `as Int?`, but Gradle `-P` values are strings, so that cast looks like it will throw.

### Scope of Impact and Testing instructions

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. CI examples are adapted from the configurations already running in this repository.

The commit includes `[skip ci]` so CI doesn't run for this docs-only change.

### Notice

> [!WARNING]
> This change must keep `main` in a shippable state; **it may be shipped without further notice**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AgUU49S4wSWd7vhJS7kEmU

PR #320, opened 2026-09-15
Add a settings reference page

DanielJette:

Settings come from several places and no page lists them together. Searches for `testinstrumentationrunnerarguments`, `adbpath` and `testclasses` return nothing.

Relates to #199

### How have you achieved it?

- New page `docs/settings.md` (sidebar position 4).
- `configure { }` settings and defaults from `TestifyConfiguration`, and the `testify { }` block from `TestifyExtension`.
- Project properties from `GradleProjectExtensions`, `ScreenshotTestTask` and `ReportTask`; instrumentation arguments from `Destination.kt` and `TestInstrumentationRegistry`; manifest tags from `TestifyFeatures`.
- Uses the real extension property `applicationPackageId`. The Gradle plugin page currently lists `targetPackageId`, which is the name printed by `testifySettings`, not a `testify { }` property.
- Notes that `outputFileNameFormat` is accepted but unused (#199). Leaves out `shardCount` / `shardIndex`, which appear to fail when passed with `-P`, and `useLocale` (#200).

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested

Every default in this page is correct — I checked all 17 `configure { }` rows and the whole `testify { }` table against `TestifyConfiguration` and `TestifyExtension`. Two changes.

### 1. The Gradle properties table is missing three real properties

The page opens with "This page lists **every** setting in each place", but the Gradle properties table omits `useLocale`, `shardCount` and `shardIndex`.

The shard pair matters: **#314 documents sharding via raw `am instrument` and the GitHub Action, but never mentions `-PshardCount`/`-PshardIndex` exist.** Please add all three rows here, and cross-link the Gradle route from #314's sharding section.

### 2. Related: `-PshardCount` was broken, and is now fixed

Worth knowing before you document it. The properties were read as:

```kotlin
shardCount = project.properties["shardCount"] as Int?
```

Gradle `-P` values always arrive as `String`, so this threw. Confirmed by running `:LegacySample:screenshotTest --dry-run -PshardCount=2 -PshardIndex=0` on `main`:

```
class java.lang.String cannot be cast to class java.lang.Integer
```

The cast only looked safe because `null as Int?` succeeds when the property is absent, which is why it went unnoticed. Fixed on branch `fix/shard-property-cast` by parsing them the way the plugin's other numeric properties already do (`device` in `AdbService`, `user` in `GradleProjectExtensions`):

```kotlin
shardCount = (project.properties["shardCount"] as? String)?.toInt()
```

Same command now succeeds; `ktlintCheck` and `test` pass on the plugin module.

### 3. `sidebar_position` collides with #319

Both pages declare 4. This one **keeps 4**; `errors.md` moves to 5, with `changelog.md` → 6 and `migration.md` → 7.

### Verified correct

- All `TestifyConfiguration` defaults, including `pauseForInspection` = 60s (`LAYOUT_INSPECTION_TIME_MS = 60000L`).
- `parallelThreads`: default `0` → CPU cores, capped at 4 — `.coerceIn(1, 4)`.
- The `isRecordMode` note that Groovy spells it `recordMode` — confirmed by `Samples/Gmd/build.gradle`.
- Annotation targets: `@ScreenshotInstrumentation` really does allow `AnnotationTarget.CLASS`, so "Test method or class" is right; `@IgnoreScreenshot` is `FUNCTION` only.
- `dev.testify.recordMode` is the literal value of `MANIFEST_IS_RECORD_MODE`.
- Issue #199 is real, open, and titled exactly "Gradle extension property outputFileNameFormat is not used".
- All instrumentation arguments, manifest tags and task names verified.

The `captureMethod` row stays as-is per discussion — effective defaults are what readers need.

AndroidTestifyBot (approved):
## Approved

All three items addressed, and the shard fix is in with a CHANGELOG entry and a version caveat I hadn't asked for.

### The properties table is now genuinely complete

I re-enumerated every `properties["…"]` read in the plugin source. The full set is exactly ten — `device`, `reportFileName`, `reportPath`, `shardCount`, `shardIndex`, `testClass`, `testName`, `useLocale`, `user`, `verbose` — and all ten are now in the table. The "lists every setting" claim in the intro holds.

`shardCount`'s "Applies to `screenshotTest` and `screenshotRecord`" is right: `InternalScreenshotTestRecordTask : ScreenshotTestTask()`, so it inherits `provideInput` and the shard handling.

### `useLocale` — citation verified

> Accepted, but the library doesn't currently use it ([#200](https://github.com/ndtp/android-testify/issues/200)).

Issue #200 is real, open, and titled exactly "Gradle project property useLocale is not used". And the claim itself holds: `Project.useLocale` in `GradleProjectExtensions.kt` has **one** definition and **zero** consumers across `Plugins/`, `Library/` and `Ext/`. `LocaleTask` doesn't read it — it only prints the device locale. (I'd assumed `LocaleTask` consumed it in my first review; the page is right and I was wrong.)

### The shard fix, verified locally

The plugin change is identical to what I'd staged. Verified on this branch rather than trusting the diff:

- `:LegacySample:screenshotTest --dry-run -PshardCount=2 -PshardIndex=0` → **BUILD SUCCESSFUL**. The same command on `main` fails with `class java.lang.String cannot be cast to class java.lang.Integer`.
- `ktlintCheck` and `test` on the plugin module → **BUILD SUCCESSFUL**.

The CHANGELOG entry and the caveat in both `settings.md` and `24-continuous-integration.md` — that these properties throw in 6.0.0 and earlier, so use the instrumentation arguments with those versions — are a better answer than just documenting the fixed behaviour. Readers on the released version get pointed somewhere that works.

The #314 cross-link closes the gap where sharding was documented only via raw `am instrument` and the GitHub Action.

### `sidebar_position`

Stays at 4; #319 moves to 5 with `changelog`/`migration` shifting to 6/7. Verified in a combined build — no ties remain anywhere in the sidebar.

### Re-verified

All 17 `configure { }` defaults, the `testify { }` table, `parallelThreads` capping via `.coerceIn(1, 4)`, `pauseForInspection` = 60s, the annotation targets, `dev.testify.recordMode`, and issue #199. Anchors `#gradle-project-properties` and `#splitting-tests-into-shards` both resolve (checked by hand — Docusaurus 2 doesn't validate anchors).

Issue #207, opened 2024-02-26
IllegalStateException: Given component holder class ComposableTestActivity does not implement interface GeneratedComponent


djette-st:
**Describe the bug**

When using HiltAndroidRule from Dagger/Hilt 2.49+ with Testify, an `IllegalStateException` is thrown.

```kotlin
@HiltAndroidTest
class NavigationTest {

    @get:Rule(order = 0)
    var hiltRule = HiltAndroidRule(this)

    @get:Rule(order = 1)
    val rule = ComposableScreenshotRule(composeTestRule = createAndroidComposeRule(ComposableTestActivity::class.java))

    @ScreenshotInstrumentation
    @Test
    fun default() {
      rule
          .setComposeActions { composeTestRule ->
              composeTestRule.setContent {
                MainNavigation()
              }
          }
          .assertSame()
    }
}
```

This issue relates to:
- [x] Compose Extension
- [ ] The Kotlin library
- [ ] The Gradle plugin
- [ ] The IntelliJ Platform plugin
- [ ] The sample code
- [ ] The documentation

**To Reproduce**

Related issue in Dagger: https://github.com/google/dagger/issues/3394

```
java.lang.IllegalStateException: Given component holder class dev.testify.ComposableTestActivity does not implement interface dagger.hilt.internal.GeneratedComponent or interface dagger.hilt.internal.GeneratedComponentManager
	at dagger.hilt.EntryPoints.get(EntryPoints.java:62)
	at dagger.hilt.android.internal.lifecycle.HiltViewModelFactory.createInternal(HiltViewModelFactory.java:206)
	at dagger.hilt.android.internal.lifecycle.HiltViewModelFactory.createInternal(HiltViewModelFactory.java:200)
	at androidx.hilt.navigation.HiltViewModelFactory.create(HiltNavBackStackEntry.kt:77)
	at androidx.hilt.navigation.compose.HiltViewModelKt.createHiltViewModelFactory(HiltViewModel.kt:57)
```

Example workaround:
https://github.com/android/nowinandroid/blob/main/ui-test-hilt-manifest/src/main/kotlin/com/google/samples/apps/nowinandroid/uitesthiltmanifest/HiltComponentActivity.kt

matteo-goghero-leitha:
I have the same problem too, event the workaround doesn't work.

DanielJette:
@matteo-goghero-leitha 

[...]

PR #321, opened 2026-09-15
Add a recipe for testing composables that use Hilt

DanielJette:

Composables that use `hiltViewModel()` fail under `ComposableTestActivity` (#207), and `hilt` returns no results in docs search.

Relates to #207

### How have you achieved it?

- New recipe `recipes/27-hilt.md`: Hilt testing dependencies, a `HiltTestApplication` runner, an `@AndroidEntryPoint` subclass of `ComposableTestActivity` declared in the debug manifest, and a test using `HiltAndroidRule` with `ComposableScreenshotScenarioRule`.
- Notes that `ComposableScreenshotRule` always launches `ComposableTestActivity`, so it can't be used with Hilt.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. The setup was run in the Flix sample (Hilt 2.60.1, KSP) on an API 37 emulator: a composable calling `hiltViewModel<NavigationViewModel>()` recorded successfully with the Hilt host activity, and the same test with `ComposableTestActivity` failed with the #207 exception. That test code was not committed.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested

Two fixes, plus a standing caveat about verification.

### 1. Step 3 contradicts itself — use the `debug` source set

> **In your `androidTest` sources**, create a subclass of `ComposableTestActivity` […]
> Declare it in the manifest of your `debug` source set, `src/debug/AndroidManifest.xml`

Those are different APKs — an `androidTest` class compiles into the test APK, while `src/debug/AndroidManifest.xml` merges into the app APK's manifest.

The `src/debug/AndroidManifest.xml` instruction is the correct half and matches the house convention (`extensions/compose/1-setup.md` and `Samples/Flix`). Please change step 3 to put `HiltComposableTestActivity` in the **`debug` source set**, not `androidTest`.

### 2. Drop the kapt fallback

> If your project uses kapt instead of KSP, use `kaptAndroidTest`.

AGP 9 removed kapt, so this is a trap rather than a fallback. Please remove the sentence. The `<hilt version>` placeholders are fine as-is — better than pinning a version that will go stale.

### Verified correct

- **`ComposableTestActivity` is `open`**, so the subclass compiles. Worth stating explicitly because the whole recipe depends on it.
- **`ComposableScreenshotRule` genuinely can't be used with Hilt** — it hardcodes `activityClass = ComposableTestActivity::class.java` in its superclass call. The `:::note` steering readers to `ComposableScreenshotScenarioRule` is correct and well-reasoned.
- `setCompose { }`, `withScenario(scenario)`, `@Rule(order = …)` and the `HiltTestRunner` snippet are all right.
- "The Testify Gradle Plugin reads this setting" — true, `TestifySettings.create` falls back to `android.defaultConfig.testInstrumentationRunner`.
- `@style/Theme.AppCompat.NoActionBar` matches what `1-setup.md` and `Samples/Flix` already use for `ComposableTestActivity`, so the Hilt host is visually consistent and existing baselines shouldn't move.
- The error string is already introduced with "an error like this", which is the right hedge for a message we can't reproduce in-repo.

### Standing caveat

This is the least-verified page in the batch: **there is no Hilt screenshot test anywhere in this repo**, so the end-to-end sequence has never been run here. The individual claims check out; the sequence has no working example behind it. Unlike #318, which lifts its examples from `FullscreenCaptureExampleTests.kt`, nothing here compiles in CI. A minimal Hilt sample would convert this from plausible to proven — worth considering as a follow-up.

AndroidTestifyBot (approved):
## Approved

Both requested changes are in, and the source-set move surfaced something I hadn't thought through.

### The contradiction is resolved

Step 3 now says "In your `debug` source set, `src/debug/java`", matching the manifest it tells you to edit and the house convention in `extensions/compose/1-setup.md` and `Samples/Flix`. The kapt sentence is gone. `<hilt version>` placeholders kept.

### The new `debugCompileOnly` section is correct

This is the part I'd have missed, and it follows directly from moving the activity out of `androidTest`:

> Code in the `debug` source set can't see your `androidTest` dependencies […]
> Both are needed, because `ComposableTestActivity` implements an interface from the core `testify` library.

Verified against the published coordinates. `ComposableTestActivity` is in `Ext/Compose`, published as `dev.testify:testify-compose` (`artifact: 'testify-compose'`), and it's declared `open class ComposableTestActivity : AppCompatActivity(), TestifyResourcesOverride` — where `TestifyResourcesOverride` lives in `Library/src/main/java/dev/testify/resources/`, published as `dev.testify:testify` (`artifact: 'testify'`). So subclassing it genuinely does need both artifacts on the compile classpath. Precisely argued.

`compileOnly` being sufficient is the right call too: the classes come from the test APK at runtime, exactly as they already do for `ComposableTestActivity` under the existing documented setup.

### Re-verified

- `ComposableTestActivity` is `open`, so the subclass compiles.
- `ComposableScreenshotRule` hardcodes `activityClass = ComposableTestActivity::class.java`, so the note steering readers to the scenario rule is correct.
- `@style/Theme.AppCompat.NoActionBar` matches what `1-setup.md` and `Samples/Flix` already use, so baselines shouldn't move.
- The error string is introduced with "an error like this" — the right hedge.

### Standing caveat, unchanged

There's still no Hilt screenshot test in the repo, so the end-to-end sequence has never been run here. Every individual claim checks out and the reasoning is sound — but this remains the one page in the batch with no executable example behind it. A minimal Hilt sample would close that gap. Approving on the strength of the reasoning and the verified parts; flagging it so it's a known, deliberate risk rather than an unexamined one.

Issue #85, opened 2022-08-05
Provide a standard way to assert multiple screenshots in a single test 


DanielJette:
Provide a way to assert multiple screenshots in a single test case. This would allow users to iterate over a list of "test cases" and generate screenshots for each test.

cc @gantonious ported from https://github.com/Shopify/android-testify/issues/126

findjigar:
This is a pretty crucial feature for us to be able to switch to this library. Hoping that this would be included soon. 🤞

DanielJette:
Hey @findjigar thanks for the feedback!

Would you mind providing more details on your requirements? I really hope to offer this feature very soon, but I would like to ensure that it meets your needs.

Do you have any thought about what the API should look like? I was thinking something like supporting multiple calls to `assertSame()` in the same test. Something like this:

```kotlin
// Take first screenshot
screenshotRule
    .setViewModifications { ... }
    .setEspressoActions { ... }
    .assertSame()

// Take second screenshot
screenshotRule
    .setViewModifications { ... }
    .setEspressoActions { ... }
    .assertSame()
```

findjigar:
@DanielJette 

Sure. We have lots of UI tests where each tests capture/compare multiple screenshots. Two primary use cases are,
- Taking screenshot of top and bottom of the screen since a lot of screens are scrollable
  - We take a top screenshot, then scroll to bottom and take another screenshot.
  - Doing this in two separate tests would take a lot of time and we would have to navigate to a particular screen again. Sometimes a screen can be 4-5 flow down the chain so we try to take multiple screenshots in single test.
- We have feature flagging for UI components, so we have tests where we would have a List of cases (each case has a unique visibility toggle), and then we would iterate on each case and take screenshot.

Another important part of this feature would be the ability to provide `unique` file name since there are multiple screenshots per test.

For example, our code looks like this,
```kotlin
class DemoTest {

     // a handy JUnit rule that stores the method name, so it can be used to generate unique
    // screenshot files per test method
    @get:Rule
    var nameRule = org.junit.rules.TestName()

    @Test
    fun screenContent_withOptional() {
        // goToParticularSreen()

        captureScreenshotAndCompare("top")
        // Scroll to Bottom
        captureScreenshotAndCompare("bottom")
    }
    
    private fun captureScreenshotAndCompare(fileNameSuffix: String) {
        val screenshotName = "${javaClass.simpleName}_${nameRule.methodName}_$name"
        // assertScreenshot(screenshotName)
    }
}
[...]

PR #322, opened 2026-09-15
Add a recipe for parameterized screenshot tests

DanielJette:

`parameterized` returns no results in docs search, and people ask how to capture several variations of a screen (#85).

Relates to #85

### How have you achieved it?

- New recipe `recipes/28-parameterized-tests.md`.
- Documents file names `Class_method[name].png` with `@Parameters(name = "{0}")`, and `Class_method[0].png` without a name, and recommends stable names.
- Points to #85 for several named screenshots in one test.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. Parameterized tests were recorded and then passed `screenshotTest` in the Legacy sample on an API 37 emulator, producing the file names shown. That test code was not committed.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Recommend closing this PR

Requesting changes to block the merge, with a recommendation to **close rather than revise**.

The page is internally accurate — I verified the baseline naming claim, and `ClientDetailsScreenshotTest_clientDetails[Alice].png` is genuinely what Testify would write, because `TestDescription.name` interpolates JUnit's `methodName` unsanitised and `Parameterized` appends the bracketed parameter name to it. The `[0]`/`[1]` fallback behaviour is right too, the warning about position-based names silently re-mapping baselines is the right thing to lead with, and issue #85 is cited correctly.

**But the underlying pattern doesn't work.** Testing found that parameterized screenshot tests only partially function, with a number of errors from the Gradle tasks in particular. Documenting them as a recommended recipe would send readers down a path that fails.

Filed as #324, which also records two source-read observations that may help a fix: the unsanitised bracket suffix in baseline names, and the fact that `DeviceUtilities.listFiles` interpolates paths straight into `adb shell ls` with no quoting, where `[` and `]` are glob characters.

Worth revisiting this page once #324 is resolved — the writing is sound and most of it would survive. The "Several screenshots in one test" section pointing at #85 is useful content that could move elsewhere in the meantime.

PR #323, opened 2026-09-15
Document right-to-left layouts in the locale recipe

DanielJette:

The locale recipe never mentions right-to-left layouts, and docs searches for `rtl` return nothing about them.

### How have you achieved it?

- Adds "Right-to-left layouts" under API 24+ in `recipes/2-locale.md`.
- Testify applies the locale with `Configuration.setLocales()`, which also sets the layout direction.
- Notes `android:supportsRtl`, that left-to-right text stays left to right, and where the baseline folder goes. Includes the import for `overrideResourceConfiguration`.

Documentation only; no library or plugin changes, so no CHANGELOG entry. Built locally with `yarn build` from `docs/`. No broken links. Both examples were recorded in the Legacy sample on an API 37 emulator: Arabic with `configure { locale }` and Hebrew with `overrideResourceConfiguration` both produced mirrored layouts in `…-ar` and `…-he` folders. That test code was not committed.

### Notice

> [!WARNING]


AndroidTestifyBot (changes requested):
## Changes requested — one fix

The RTL content is correct. One change, and it's caused by this PR landing rather than by anything wrong in it.

### Also remove `launchActivity = false` from the existing example

The new `ScreenshotRule` example correctly omits `launchActivity`:

```kotlin
@get:Rule var rule = ScreenshotRule(
    activityClass = TestLocaleHarnessActivity::class.java,
    rootViewId = R.id.harness_root
)
```

That's right — the primary constructor has no such parameter (it passes `false` to `ActivityTestRule` unconditionally), and the overload that still accepts it is `@Deprecated("Parameter launchActivity is deprecated and no longer required")`.

But the **pre-existing** example a few paragraphs above still passes `launchActivity = false`, so once this merges the page shows both forms with no indication which to copy. Please drop it from the older example too. Current docs only — no need to touch `versioned_docs`.

### Verified correct

- **`37-1080x2220@440dp-ar` is the right folder shape.** `DEFAULT_FOLDER_FORMAT` is `"a-wxh@d-l"` and the locale segment comes from `toLanguageTag().replace("-", "_")`, which for `Locale.forLanguageTag("ar")` is just `ar` — so correctly no country segment, unlike the `fr_FR` and `ja_JP` folders in `Samples/Legacy`. Easy to get wrong.
- `overrideResourceConfiguration<T>(locale = …)` exists and does live in `dev.testify.internal.helpers`, so the shown import is accurate.
- The scenario-rule vs `configure { }` split matches `NoResourceConfigurationOnScenarioException`.
- `android:supportsRtl="true"` and the locale-derived layout direction are correct platform behaviour, and the caveat about LTR placeholder text still laying out LTR is a genuinely useful thing to say.

### Follow-up (not blocking)

`Samples/Legacy` has `en_US`, `en_CA`, `fr_FR` and `ja_JP` but no `ar` or `he`, so RTL isn't exercised anywhere in the repo. Adding one `he` or `ar` baseline to the Legacy sample would make this page self-evidently true and keep it honest in CI.

### Minor note

`ScreenshotScenarioRule(rootViewId = R.id.harness_root)` in the new example is arguably more correct for a harness activity than the bare `ScreenshotScenarioRule()` above it — suggest keeping the new style. Separately, surfacing `dev.testify.internal.helpers` in user-facing docs is a little awkward; the function is public so it works, but relocating it out of an `internal`-named package would be a nicer long-term answer.

AndroidTestifyBot (approved):
## Approved

`launchActivity = false` is gone from the pre-existing `ScreenshotRule` example, so the page now shows one consistent constructor form throughout. That was the only blocking item.

The rest of the section was verified in the previous round and is unchanged:

- `37-1080x2220@440dp-ar` is the correct folder shape — `DEFAULT_FOLDER_FORMAT` is `"a-wxh@d-l"`, and the locale segment is `toLanguageTag().replace("-", "_")`, which for `Locale.forLanguageTag("ar")` yields bare `ar` with no country segment. Correctly different in shape from the `fr_FR` and `ja_JP` folders in `Samples/Legacy`.
- `overrideResourceConfiguration<T>(locale = …)` exists and does live in `dev.testify.internal.helpers`.
- The scenario-rule vs `configure { }` split matches `NoResourceConfigurationOnScenarioException`.
- `android:supportsRtl="true"` and the locale-derived layout direction are correct, and the note that LTR placeholder text still lays out LTR is a genuinely useful caveat.

Docs build verified locally: exit 0, zero warnings, individually and combined with the other open PRs.

### Follow-up, as agreed

`Samples/Legacy` still has no `ar` or `he` baseline, so RTL isn't exercised in the sample suite. Worth adding one so this page stays honest in CI — not blocking.

